PR #31, opened 2018-09-24
Handle connection interruptions by exiting with failure

mjkoo:
Related to polyswarm/ambassador#1 (thanks @strazzere!) , currently we've decided its better to exit on error while we come up with a better recovery strategy given potentially stale state in the client.


strazzere:
👍 this is a much better looking solution that the hack I had!

PR #37, opened 2018-10-02
Vartifacts

scsich:
Replace `withartifacts` docker tag with a Pythonic puller that downloads a basic corpus to arbiter and ambassador. Tested in staging.

### DIFF
--- a/src/corpus/__init__.py
+++ b/src/corpus/__init__.py
@@ -3,10 +3,10 @@ import tempfile
 import urllib.request
 import logging
 from subprocess import check_call
-from arbiter.verbatimdb.__main__ import generate_db
+from arbiter.verbatimdb.db import generate_db
 
 MALICIOUS_BOOTSTRAP_URL = os.getenv('MALICIOUS_BOOTSTRAP_URL')
-ARCHIVE_PW = os.getenv('ARCHIVE_PASSWORD')
+ARCHIVE_PASSWORD = os.getenv('ARCHIVE_PASSWORD')
 
 
 class DownloadToFileSystemCorpus(object):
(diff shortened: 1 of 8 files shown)


src/corpus/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+        self.base_dir = base_dir if base_dir else tempfile.mkdtemp(suffix="malware-corpus")
+        self.truth_fname = "truth.db"
+        self.truth_db_pth = os.path.join(self.base_dir, self.truth_fname)
+
+    @property
+    def mal_path(self):
+        return os.path.join(self.base_dir, "malicious")
+
+    @property
+    def benign_pth(self):
```
mjkoo: s/pth/path/ or mal_path above should be made consistent

Edit: really all instances of pth and path

src/corpus/__init__.py
```diff
@@ -0,0 +1,65 @@
+import os
+import tempfile
+import urllib.request
+import logging
+from subprocess import check_call
+from arbiter.verbatimdb.__main__ import generate_db
+
+MALICIOUS_BOOTSTRAP_URL = os.getenv('MALICIOUS_BOOTSTRAP_URL')
+ARCHIVE_PW = os.getenv('ARCHIVE_PASSWORD')
```
mjkoo: s/ARCHIVE_PW/ARCHIVE_PASSWORD/ imo

src/corpus/__init__.py
```diff
@@ -0,0 +1,65 @@
+import os
+import tempfile
+import urllib.request
+import logging
+from subprocess import check_call
+from arbiter.verbatimdb.__main__ import generate_db
```
mjkoo: Probably should pull this out of __main__ if we're using it in other packages

src/corpus/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+        self.truth_fname = "truth.db"
+        self.truth_db_pth = os.path.join(self.base_dir, self.truth_fname)
+
+    @property
+    def mal_path(self):
+        return os.path.join(self.base_dir, "malicious")
+
+    @property
+    def benign_pth(self):
+        return os.path.join(self.base_dir, "malicious")
```
mjkoo: this looks like a copy paste error, should be "benign"

src/corpus/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def download_and_unpack(self):
+        logging.info("Downloading to {0}".format(self.base_dir))
+        for is_mal, package_name in [(True, "malicious.tgz.gpg"), (False, "benign.tgz.gpg")]:
+
+            mal_dir = self.mal_path if is_mal else self.benign_pth
+            archive_pth = os.path.join(mal_dir, package_name)
+            archive_dst = mal_dir
+            if not os.path.isdir(archive_dst):
+                os.mkdir(archive_dst)
```
mjkoo: Think this will fail if multiple dirs need to be created (not equiv to `mkdir -p`) os.makedirs does do this.

src/corpus/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+    def get_benign_file_list(self):
+        return self._get_pth_listing(self.benign_pth)
+
+    def download_truth(self):
+        u = "{0}/{1}".format(self.url, self.truth_fname)
+        logging.info("Fetching truth database {0}".format(u))
+        r = urllib.request.urlretrieve(u, self.truth_db_pth)
+
+    def generate_truth(self):
+        d = generate_db(self.truth_db_pth, self.mal_path, self.benign_pth)
```
mjkoo: `generate_db` doesn't return anything, and we dont use d at all

src/corpus/test_dl.py
```diff
@@ -0,0 +1,21 @@
+import os
+import unittest
+from corpus import DownloadToFileSystemCorpus
+
+
+class DownloaderUnitTest(unittest.TestCase):
```
mjkoo: Tests depend on env variable and live server, would be nice if they didnt

scsich: Yeah, this is a larger question if we want to expose a test corpus location. Leaving for now.

Issue #42, opened 2018-10-07
IPFS error causes ambassador to freeze


mjkoo:
We should:

- [x] Try more than once if IPFS artifact retrieval error happens
- [x] If one bounty fails to submit, log error and proceed to next
- [ ] If N block pass without a bounty submission, log error and die

mjkoo:
Closed by #43

PR #43, opened 2018-10-07
Fix error where failed tx-generating request will hold the nonce lock…

mjkoo:
…, deadlocking the application

ref #42


scsich:
LGTM, but let's make a note on request timeouts being configurable when we redo configuration here and with polyswarmd.

mjkoo:
ack, yeah were using the default aiohttp timeout of 5min before which seemed excessive

PR #50, opened 2018-10-12
Add a relay driver to polyswarm-client

Rizato:
Relay can be used from the command line to deposit/withdraw funds from a sidechain. 

Added relayclient and balances client to polyswarm-client to handle the polyswarmd specifics of getting a balance and depositing/withdrawing funds.


Rizato:
We probably don't want to merge this until relay e2e has been added.

Rizato:
Per #55 This needs to use the same nonce recovery

Rizato:
@mjkoo It is just a one time transfer to/from relay. 

The only reason I called it backend, and kept it as an optional arg was for consistency with the other drivers. Removing it, and just doing a positional arg like the deposit amount would be better too.

mjkoo:
That makes sense, but i wonder if we could instead do 

1) a relay client that just handles deposits/withdrawals (no backend, simple cli way to interact)
2) a 'balance' driver that has this balance watching and top-off kind of functionality

kind of matching the split between relayclient and balanceclient for instace?

PR #53, opened 2018-10-15
Fix/bump version

Rizato:
Pip won't pull the changes adding guid to settle events because polyswarm-client didn't have an updated version.


Rizato:
This did not cause issues with our docker-hub versions because the envs are fully rebuilt each time. Only caused issues for local environments.

Issue #52, opened 2018-10-15
Resync nonce


mjkoo:
**Describe the bug**
When nonce gets out of sync ('nonce too low' error from polyswarmd) we should try to grab the nonce lock and re-fetch the correct nonce from polyswarmd instead of giving up. Ideally this does not happen but the situation can arise outside of our control due to sending transactions from e.g. two clients.

**To Reproduce**
1. Send transaction outside of running polyswarm-client with same key
2. Watch nonce errors

**Expected behavior**
On nonce error, attempt to resync nonce

**Additional context**
Occasionally seen in logs, not sure what is causing desync

myhndl:
Example: https://app.datadoghq.com/logs?cols=%5B%22core_host%22%2C%22core_service%22%5D&context_event=AAAAAWZ4mIsX3hvYVgAAAABBV1o0bUlzWXllb1YtbHN4VkhSaw&event=AAAAAWZ4mIsX3hvYVgAAAABBV1o0bUlzWXllb1YtbHN4VkhSaw&from_ts=1539536148384&index=main&live=false&messageDisplay=inline&query=host%3Ai-0801d251bb87abd2e+service%3Adocker+container_id%3A2522899285e45f75d014dc0ca85598ec5a498154e6173847c322225cb35fa072&stream_sort=desc&to_ts=1539621555153

myhndl:
Is this addressed by this: https://github.com/polyswarm/polyswarm-client/pull/55 @mjkoo

mjkoo:
Yes closed

PR #55, opened 2018-10-16
Resync nonces if transaction errors are occurring

mjkoo:
Addresses #52


mjkoo:
Tested this locally by running two concurrent ambassadors to trigger the fail condition, seemed to sort itself out.

PR #64, opened 2018-10-24
Exceptions should not be swallowed by naked-except blocks

zv:


src/polyswarmclient/__init__.py
```diff
@@ -47,11 +47,11 @@ def is_valid_ipfs_uri(ipfs_uri):
...
-    except Exception:
-        pass
-
-    logger.error('Invalid IPFS URI: %s', ipfs_uri)
-    return False
+    except TypeError:
+        logger.error('Invalid IPFS URI: %s', ipfs_uri)
+        return False
+    except Exception as err:
+        logger.exception('Unexpected error: %s', err)
```
myhndl: This case needs a `return False`, right?

myhndl: Or better: `return False` outside of the exception clauses.

Now I see what you were asking about before.

zv: @myhndl My idea here is that we should bubble up the exception because we've encountered a condition that is _more_ exceptional than a mangled `ipfs_uri`.

zv: @myhndl Thank you! Changes made.

PR #74, opened 2018-10-31
Feature/unit tests

mjkoo:
Flesh out unit test suite


tests/test_corpus.py
```diff
@@ -0,0 +1,23 @@
...
+    pass
+
+    #d = DownloadToFileSystemCorpus()
+    #d.download_and_unpack()
+
+    #assert d.get_malicious_file_list()
+    #assert d.get_benign_file_list()
+
+    #d.generate_truth()
+    #assert os.path.exists(d.truth_db_pth)
```
ceseale: Should these still be commented out?

ceseale:
Great job! Look good overall :+1:

tests/test_corpus.py
mjkoo: Yeah don't want unit tests depending on external services, have a separate ticket about corpus stuff will add in "Restore test cases but with mocked responses" to that todo list

zv: I think @mjkoo has it on the nose, no unit tests should communicate with a remote service.

I moved these outside of `corpus` and got them running only to get started on the unit test work (they were the only ones in the project at the time).

PR #80, opened 2018-11-01
Use extra for dicts

### LINKED ISSUE #79: Logs that contain JSON objects, should log fully as JSON
**Is your feature request related to a problem? Please describe.**
We print of a lot of logs similar to the following `logger.info('Received settled bounty on chain %s: %s', chain, data)`
We end up having logs where the inner JSON is not parsable by datadog per https://github.com/polyswarm/development-private/issues/73

**Describe the solution you'd like**
Add wrap the `data` json object inside a new one containing the message and chain

Rizato:
Adds the python dict output into logs as `extra=`

When using logformat json, it will separate out the dict so it the dict can be parsed as json.
When using logformat text, it will appear the same as before.

Also, I changed up the drivers to use `--log-format` instead of `--log_format`

Fixes #79


src/polyswarmclient/config.py
```diff
@@ -3,6 +3,29 @@
 from polyswarmclient.log_formatter import JSONFormatter
 from pythonjsonlogger import jsonlogger
 
+EXTRA_KEYS = ['assertion', 'bounty', 'channel', 'errors', 'quorum', 'response', 'results', 'reveal', 'settle', 'verdict']
```
mjkoo: Is there a way to not have to enumerate all these names in here? Seems like an extra place to need to make a modification in future just to print a new log that doesnt match existing types

Rizato: The easiest way would be to use a key of default terms that logging looks for and blacklist them. Then just add everything else we find.

mjkoo: hmm messy on python's side that these just get added to the __dict__ of the logrecord imo... blacklisting seems kinda fragile, how about wrapping stuff into e.g. `{'extra': blah}` and just looking for the extra key? could write a small helper for that or something perhaps

PR #95, opened 2018-11-19
Remove default chain values

### LINKED ISSUE #87: Remove or think critically about default parameters
**Describe the bug**
Notably `chain='home'` everywhere, this has hidden a bug or two and isn't even a good assumption anymore (at one point it was, but no longer)

### LINKED ISSUE #94: Switch CLI chains default to side
**Is your feature request related to a problem? Please describe.**
With relay up, we are moving everything to the sidechain. Having the default set to homechain has tripped things up a few times. 

**Describe the solution you'd like**
Set default to sidechain

**Additional context**
Similar to #87

Rizato:
Switch cli commands to sidechain
Remove default chain='home' values and the like

Fixes #87 
Fixes #94


Rizato:
Gotta remove that vscode file. One moment.

PR #107, opened 2018-12-06
preparing for PyPI publishing.

mrtizmoatwork:
only thing left to do is to add a check to ensure that the git tag == setup.py version string.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -3,13 +3,12 @@ image: docker:stable
 services:
   - docker:dind
 
-
-
 stages:
   - test
   - e2e
   - prebuildwindows
   - notifydependents
+  - pypi
 
 test:
   image: python:3.5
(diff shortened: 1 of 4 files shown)


mrtizmoatwork:
Important note @myhndl when tagging a release, we need to annotate the tag for this to work optimally.

The command used to tag the branch should be like this:

`git tag -a -s -m "release 0.2.4" 0.2.4`

That is an annotated tag that is signed with an annotation message.

-m: message
-a: annotated
-s: signed

mrtizmoatwork:
Tagging via the GitHub UI should also create an annotated tag if you include a message.

myhndl:
Wouldn't it be better to just `git describe --tags`

mrtizmoatwork:
you can choose which ever you want:

Here is the most useful link I found:
https://stackoverflow.com/questions/1404796/how-to-get-the-latest-tag-name-in-current-branch-in-git

mrtizmoatwork:
Important quote "Unless the tags are annotated tags, I think there is no way for git to distinguish which of the two was created earlier. " on that page.

myhndl:
Ah gotcha, okay annotated it is!

PR #130, opened 2018-12-19
Clamav use async socket

mjkoo:


requirements.txt
```diff
@@ -3,7 +3,7 @@ aiohttp==2.3.1
 aioresponses==0.5.0
 asynctest==0.12.2
 base58==0.2.5
-clamd==1.0.2
+git+https://github.com/polyswarm/python-clamd.git@async#egg=clamd
```
mrtizmoatwork: why are we hosting a fork of clamd? Remember, this clamav engine is used in docs.polyswarm.io as an example, so we need to make sure that others doing our examples can still follow along.

mrtizmoatwork: Oh. I see you created a new branch for async. Ok. Then we need to make a note to add info about async to our docs.

setup.py
```diff
@@ -5,7 +5,7 @@
 
 def parse_requirements():
     with open('requirements.txt', 'r') as f:
-        return f.read().splitlines()
+        return [r for r in f.read().splitlines() if not r.startswith('git')]
```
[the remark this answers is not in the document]
mjkoo: git requirement breaks this parse_requirements logic and theres no good way to specify dep on a git repo in setup.py afaik, will look at opening a PR to upstream

PR #134, opened 2018-12-26
add clamav arbiter

ceseale:
adds clamav arbiter related to https://github.com/polyswarm/development-private/issues/115


ceseale:
I agree, I would have preferred to do something similar to https://github.com/polyswarm/engine-k7/pull/6. But I figured it was better to use the same pattern to avoid altering the docs/microengine cli command.

PR #135, opened 2018-12-28
Fix 'Future attached to different loop' error

zv:


src/polyswarmclient/__init__.py
```diff
@@ -150,7 +150,8 @@ def run(self, chains=None):
         # Default event loop does not support pipes on Windows
         if sys.platform == 'win32':
             loop = asyncio.ProactorEventLoop()
-            asyncio.set_event_loop(loop)
+            asyncio.get_event_loop_policy().set_event_loop(loop)
```
mjkoo: Why change this

zv: The docs claim `set_event_loop` is equivalent but I got an error about complaining about invalid state when I was working with Zillya earlier.

src/polyswarmclient/__init__.py
```diff
@@ -150,7 +150,8 @@ def run(self, chains=None):
...
+            self.base_nonce_lock = {'home': asyncio.Lock(), 'side': asyncio.Lock()}
```
mrtizmoatwork: I think this should be all the way left aligned, not under the "if" test. AND then I think the initial definition of self.base_nonce_lock should be = {'home': None, 'side': None} OR = {}.

The whole problem is caused because we take something from the default event loop and save it to a variable prior to changing the event loop. So, if we remove the code in the __init__ that saves something from the default eventloop and put it here, immediately after where we test/change the event loop, we should solve the problem in a clean way.

I'm going to try what I've stated above and see if it works. I'll post a commit here if it does.

mrtizmoatwork:
@mjkoo @zv I tested the hypothesis that I proposed above using the xvirus engine on Windows.

1. I removed the `.get_event_loop_policy()` call that @zv added and re-set it to the original form
2. I removed lock() instantiation from the `__init__` method
3. I shifted the lock instantiation that @zv added to not be inside the `if` tests. That way it always runs.

My changes are in commit `aa911c6` above. @zv please test them on your other engines.

src/polyswarmclient/__init__.py
```diff
@@ -152,6 +152,7 @@ def run(self, chains=None):
             loop = asyncio.ProactorEventLoop()
             asyncio.set_event_loop(loop)
 
+        self.base_nonce_lock = {'home': asyncio.Lock(), 'side': asyncio.Lock()}
```
mjkoo: I like this better, but run_task itself is also exposed so as to allow users to override what loop they're using explicitly, this way the locks wont be initialized. Think we should move this to run_task as there we assume that whatever loop we're running on is already set up.

mrtizmoatwork: ok. changing that now.

mrtizmoatwork:
tested these latest changes with xvirus. everything looks good.

PR #136, opened 2018-12-28
Don't trust polyswarmd

Rizato:
Need to refactor into cleaner OOP. 
Also, need to standardize logging.


src/polyswarmclient/verify.py
```diff
@@ -0,0 +1,412 @@
...
+        self.w3 = Web3()
+        self.account = account
+
+    def bool_list_to_int(self, bool_list):
+        retval = 0
+        for b in bool_list:
+            retval << 1
+            retval += 1 if b else 0
+
+        return retval
```
mrtizmoatwork: is length of bool_list always going to be <= the number of bits in an Int? Do we care if it goes longer?

src/polyswarmclient/verify.py
```diff
@@ -0,0 +1,412 @@
...
+        value = transaction.get('value')
+        if transaction_data is None or transaction_to is None or value is None:
+            return None
+        byte_data = bytes(HexBytes(transaction_data))
+        sig = byte_data[:4]
+        data = bytes(HexBytes(byte_data[4:]))
+        to = w3.toChecksumAddress(transaction_to)
+        return cls(to, value, sig, data)
+
+class AbstractTransactionVerifier():
```
mrtizmoatwork: this class should inherit from 'ABC'
https://docs.python.org/3/library/abc.html

src/polyswarmclient/verify.py
```diff
@@ -0,0 +1,412 @@
...
+                data = decode_single(abi[0], transaction.data)
+            else:
+                data = decode_abi(abi, transaction.data)
+        except InsufficientDataBytes:
+            logger.error('Transaction did not match expected input data')
+            return False
+
+        return self.verify_transaction(transaction.to, transaction.value, transaction.signature_hash, data)
+
+    def verify_transaction(self, data_tuple, transaction):
```
mrtizmoatwork: decorate with `@abstractmethod`

src/polyswarmclient/verify.py
```diff
@@ -0,0 +1,412 @@
...
+        This function will verify the transactions, and determines if the transactions are expected.
+
+        Args:
+            transactions (list): Transactions to be verified before signing and returning to polyswarmd
+        Returns:
+            True if valid and expected
+        """
+        raise NotImplementedError('Verify is not implemented')
+
+    def get_abi(self):
```
mrtizmoatwork: decorate with `@abstractmethod`

src/polyswarmclient/verify.py
mrtizmoatwork: ok

PR #137, opened 2018-12-28
bump ver to 1.0rc5

mrtizmoatwork:
* updated version to 1.0rc5. 
* filled out major changes for each release in the HISTORY.md file.

We should add any remaining changes to the HISTORY.md before we cut rc5.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,41 +1,42 @@
 # Release History
 
-### 1.0rc5 (2018-12-XX)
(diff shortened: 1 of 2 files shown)


Rizato:
I have another big change coming in today.

Rizato:
Shortly

mrtizmoatwork:
No problem. I just wanted to get the HISTORY updated and added for the next release. We can cut the release when the time is right. But today sometime would be ideal.

PR #139, opened 2018-12-30
fixed use of asyncio.Lock() in balancemanger.Maintainer()

mrtizmoatwork:
* fixed another case of asyncio.Lock() being used before the loop is changed. changed it to create the locks after the client is run using run_on.register().


mrtizmoatwork:
@Rizato @mjkoo why would there be an exception about a NoneType chain when the chains dict is set to be home and side?

Here is the test that is failing: https://gitlab.polyswarm.io/externalci/e2e/-/jobs/11245
It has the output I posted in the previous post.

mrtizmoatwork:
woot. that fixed it.

Issue #138, opened 2018-12-28
Remove Default Backend


myhndl:
**Is your feature request related to a problem? Please describe.**
Currently we default to the `scratch` backend. This may mislead people as to the health of their engine.

**Describe the solution you'd like**
Remove default backend, fail hard if none specified.

PR #148, opened 2019-01-03
Remove default backend of 'scratch'.

zv:
I added an exception at the beginning of `choose_backend` to make the error explicit, rather than relying on the existing exceptions later in the function.


zv:
Resolves #138

myhndl:
Might even be sufficient to not specify a default.

zv:
@myhndl Pushed the requested changes, ran tests and verified that `click` doesn't erroneously break on `required=True` where only an environment variable is passed.

Issue #144, opened 2019-01-02
Exits on Unhandled Exception


myhndl:
**Describe the bug**
We're exiting when an unhandled exception occurs, e.g. a timeout.

It happening on Tachyon:
https://app.datadoghq.com/logs?cols=%5B%22core_host%22%2C%22core_service%22%2C%22log_http.status_code%22%2C%22log_request_Authorization%22%5D&context_event=AAAAAWgN3Mowp4_NmgAAAABBV2dOM01wQ1hvTjBBelpqc1d1Wg&event=AAAAAWgN3Mowp4_NjQAAAABBV2dOM01wQlhvTjBBelpqc1d1Sw&from_ts=1546280108117&index=main&live=false&logset=main&messageDisplay=inline&query=host%3AEC2AMAZ-DALA5CT+service%3Amicroengine+filename%3Aengine.log&saved_view&stream_sort=desc&to_ts=1546452531421

**To Reproduce**
Run a microengine for a while.

**Expected behavior**
We should not exit; polyswarm-client must do everything to remain available.

myhndl:
Also affecting `consumer_backend`:
```
user@user-VirtualBox:~/work/infrastructure$ docker service logs -t --tail 3 pgzbblzj9o2a
2019-01-02T20:20:23.697482206Z consumer_backend.1.pgzbblzj9o2a@ip-172-31-82-214    | {"level": "INFO", "name": "consumer.ambassador", "timestamp": "2019-01-02T20:20:23.697257Z", "message": "Vote received for bounty c645e5a1-d71b-4568-8f2a-4c40d71b0da5 0xd8b48Da78188312c5fC079E532afd48De973767E: [False]"}
2019-01-02T20:20:23.698995015Z consumer_backend.1.pgzbblzj9o2a@ip-172-31-82-214    | {"level": "INFO", "name": "consumer.ambassador", "timestamp": "2019-01-02T20:20:23.698635Z", "message": "Vote on file c03d23b734567a3deba88fc58cf721f5233e9c958f03268d72984a2bf9ac48a5: False"}
2019-01-02T20:20:23.718341227Z consumer_backend.1.pgzbblzj9o2a@ip-172-31-82-214    | {"level": "ERROR", "name": "polyswarmclient", "timestamp": "2019-01-02T20:20:23.718108Z", "message": "Detected unhandled exception, exiting with failure"}
```

myhndl:
Also affecting `ambassador`:
```
[...]

PR #152, opened 2019-01-03
Fix/recover from unhandled exception

mjkoo:
Ref #144 and #146 , think we should still handle known error cases locally when we can, such as retrying websocket connection inside event task, but this changes top level error handler to start up a new main task on exit


mjkoo:
Builds on #151 supersedes #146

mjkoo:
Tested by throwing an exception if tries < N in main task after init, recovers after that

src/balancemanager/__init__.py
```diff
@@ -71,7 +72,6 @@ def run(self):
                             self.client.from_wei(balance), self.client.from_wei(amount_wei))
 
             self.exit_code = 1
```
Rizato: Does this exit_code actually do anything?

PR #155, opened 2019-01-03
Exit Client with `os.exit()' when sys.platform == 'win32'

zv:
This code calls `os._exit` which is a hard exit. No exception handlers/destructors/exit routines are run, threads die, etc.

This begs the question:

Can we handle clean-up manually *prior* to calling `sys.exit`?

Can we manually see what exception handler *would* be called if an exception *were* to be invoked (possibly allowing us to fix it)? If it's not an exception handler issue - is it outstanding threads?

[_voice grows increasingly insane_] Can we reach into a python coredump and reconcile regular python pdb symbols with interpreter state to reconstruct where in the application it is hanging at a "high level"?

The possibilities for how to handle this are endless but this code here does have the advantage of working now and showing no clear defects.

### DIFF
--- a/src/polyswarmclient/__init__.py
+++ b/src/polyswarmclient/__init__.py
@@ -159,7 +159,7 @@ class Client(object):
         if self.exit_code:
             self.exit_loop(self.exit_code)
 
-    def exit_loop(exit_status):
+    def exit_loop(self, exit_status):
         """
         Exit the program entirely.
         """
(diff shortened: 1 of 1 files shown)


scsich:
@zv extract the exit_loop inner function to a class method `self.exit_loop` so someone can override by subclassing Client. Other than that LGTM

PR #158, opened 2019-01-04
Updated HISTORY.md with 1.0.1 changes

zv:


Rizato:
I think there are a few more changes in master right now. 

Also, I’m not sure how we are doing versioning, but the ambassador loop changes are breaking.

PR #164, opened 2019-01-10
Pass mask, verdicts, and metadata to bid.

Rizato:
~Wip because I need to test it~ Works fine


mjkoo:
Note that this is a breaking change affecting microengines, lgtm though

PR #169, opened 2019-01-21
Change nonce handling for minimal lock time & faster throughput

Rizato:
Ignore polyswarmd's handling of nonces with the base_nonce value on transaction building requests.
Instead, sends 0 every time, and replaces the nonce in each transaction to the correct value.
This reduces the amount of time that is guarded by the nonce lock.

The above results in a significantly faster response time for bounties & does not have the same buildup as transactions increase.

Also, removed the limit on open connections in aiohttp, because once settles get added to the transactions, the limit started slowing things down.  

Changed up the update nonce behavior. It no longer locks and updates, as that lock wouldn't be acquired for some time, given all the other transactions already waiting on the nonce. Instead, it sets a flag to true, and the next time the lock is acquired, the nonce will be updated.

Part of the solution for polyswarm/development-private#159


Rizato:
While testing you can expect to see inactive bounties on large files, which is a different issue not related to the nonce stuff that we have to solve. Otherwise, it should function just fine.
Also, nonce too low and subsequent recoveries are common. 

In addition to normal tests, run the scratch microengine against stage & monitor for `bounty inactive` errors on transactions. (I just added my stage key to the docker image & ran against stage)

src/polyswarmclient/__init__.py
```diff
@@ -382,11 +431,15 @@ def exit(self, exit_status):
...
+
+            nonce_manager = self.base_nonce[chain]
+            nonces = await nonce_manager.next(amount=len(transactions))
+            transactions = [self.replace_nonce(nonces[i], transaction) for i, transaction in enumerate(transactions)]
             tx_result = await self.post_transactions(transactions, chain, api_key=api_key)
             errors = tx_result.get('errors', [])
             nonce_error = False
             for e in errors:
-                if 'invalid transaction error' in e.lower():
+                if 'known transaction' in e.lower() or 'nonce too low' in e.lower() or 'timeout waiting on receipt' in e.lower() or 'replacement transaction underpriced' in e.lower():
```
Rizato: This is a mistake here. Done for testing. Seeing if I could skip some of them. I think I ended up with each added in manually.

src/polyswarmclient/__init__.py
```diff
@@ -27,6 +28,61 @@
...
+
+    async def acquire(self):
+        """
+        Acquires the nonce lock.
+        Updates base_nonce if needs_update is set
+        """
+        await self.lock.acquire()
+        if self.needs_update:
+            # Small delay to allow in progress transactions to return since they will all trigger this too
+            await asyncio.sleep(1)
```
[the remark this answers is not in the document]
Rizato: The sleep is just the random time that worked best. It would be better if I could wait for all the submitted transactions to return directly. Not sure how to modify it to do that.

src/polyswarmclient/__init__.py
```diff
@@ -27,6 +27,91 @@
 REQUEST_TIMEOUT = 300.0
 
 
+class NonceManager:
+    """
+    Managers the nonce for some ethereum chain
```
mjkoo: manages

src/polyswarmclient/__init__.py
```diff
@@ -27,6 +27,91 @@
...
+    """
+    def __init__(self, client, chain):
+        self.base_nonce = 0
+        self.client = client
+        self.chain = chain
+        self.needs_update = asyncio.Event()
```
mjkoo: If just checking that this is set but never wait()-ing on it, might as well be a boolean

src/polyswarmclient/__init__.py
```diff
@@ -27,6 +27,91 @@
...
+                self.in_progress += 1
+        return results
+
+    async def finished(self):
+        """
+        Mark that some in-progress transactions finished on polyswarmd
+        """
+        async with self.progress_lock:
+            self.in_progress -= 1
+        self.in_progress_event.set()
```
mjkoo: So in_progress_event is set every time a tx is finished, but is only cleared when wait_all_finish is called after needs_update logic is triggered, if the event is set wait returns immediately, in the common case I dont think this will actually wait for an event in wait_all_finished()?

tests/test_client.py
```diff
@@ -29,12 +30,20 @@ def test_is_valid_ipfs_uri():
 @pytest.mark.asyncio
 async def test_update_base_nonce(mock_client):
     mock_client.http_mock.get(mock_client.url_with_parameters('/nonce', chain='home'), body=success(42))
-    await mock_client.update_base_nonce(chain='home')
-    assert mock_client.base_nonce['home'] == 42
+    home = polyswarmclient.NonceManager(functools.partial(mock_client.update_base_nonce, 'home'))
```
mjkoo: Think these need updates to new constructor

PR #173, opened 2019-01-23
Remove received event logs and update to 1.0.3

Rizato:
Reducing datadog messages


Rizato:
We use these values. Need to make it optional instead.

PR #182, opened 2019-02-05
Use split /transactions route

Rizato:
Use the split /transactions route for getting events from a transaction. 

Also, does verification of the result to handle the case where some transactions fail, but other succeed. 

Refactored into Transaction objects that handle verification of the transactions before signing, and making sure the desired event came back.


Rizato:
For testing this, I an orchestration and did two things to trigger errors. Ran microengines with `docker-compose -f base.yml -f clients.yml up --scale microengine=2 --no-deps microengine` which will spawn two microengines using the same key. 

Also, just for general nonce handling, did the deposit-microengine but withdraws 2 instead of depositing.

Another way to test, is to modify post_transactions to call make_request twice for building the transaction.

Rizato:
```
success, results = await self.make_request('POST', '/transactions', chain, json={'transactions': signed},
                                                     api_key=api_key, tries=1)
        success, results = await self.make_request('POST', '/transactions', chain, json={'transactions': signed},
                                                     api_key=api_key, tries=1)
```

Will do the trick. If you add some logging you will see that it always gets known transaction the second time, but still does the reveals and settles.

PR #183, opened 2019-02-05
Paramters object allowing updates after first run

mjkoo:
Breaking change


src/polyswarmclient/bountiesclient.py
```diff
@@ -107,7 +110,7 @@ def __init__(self, client):
             'uri': artifact_uri,
             'duration': duration,
         }
-        bounty_fee = self.parameters[chain]['bounty_fee']
+        bounty_fee = self.parameters[chain].get('bounty_fee')
```
Rizato: Should this be awaited?

src/polyswarmclient/bountiesclient.py
```diff
@@ -160,7 +163,7 @@ def __init__(self, client):
             'mask': mask,
             'commitment': commitment,
         }
-        fee = self.parameters[chain]['assertion_fee']
+        fee = self.parameters[chain].get('assertion_fee')
```
Rizato: Same with this one

src/polyswarmclient/bountiesclient.py
mjkoo: ya ty

src/polyswarmclient/bountiesclient.py
```diff
@@ -67,7 +68,7 @@ def has_required_event(self, transaction_events):
         for assertion in assertions:
             if (assertion.get('bid', 0) == self.bid and
                     assertion.get('mask', []) == self.mask and
-                    assertion.get('nonce', '') == self.nonce and
+                    assertion.get('commitment', '') == self.commitment and
```
mjkoo: @Rizato rebased off current master, in resolving changed this to commitment, pretty sure thats what we want here. Thoughts?

Rizato: Yea, that does make sense.

PR #184, opened 2019-02-06
Feature/redis microengine

mjkoo:
Builds on #183


mjkoo:
Test in orchestration with https://github.com/polyswarm/orchestration/blob/feature/redis-microengine/producerconsumer.yml

PR #186, opened 2019-02-08
Pad bool list with False to given length

Rizato:


src/polyswarmclient/utils.py
```diff
@@ -16,9 +19,19 @@ def bool_list_to_int(bs):
...
+def int_to_bool_list(i, expected_size):
+    # return empty list when 0 and no items expected (Return actual value if > 0)
+    if expected_size == 0 and i == 0:
+        return []
+
     s = format(i, 'b')
-    return [x == '1' for x in s[::-1]]
+    bool_list = [x == '1' for x in s[::-1]]
+    diff = expected_size - len(bool_list)
+    if diff > 0:
```
mjkoo: Think this is overly defensive, :man_shrugging:

PR #190, opened 2019-02-12
Feature/microengine loggerbuilder

zv:


mrtizmoatwork:
https://github.com/polyswarm/polyswarm-client/blob/master/src/polyswarmclient/config.py

^ this already exists. Use it instead.

PR #193, opened 2019-02-13
Fix/get tx timeouts

mjkoo:
Fixes various issues with nonce sync + tx validation

Currently on staging


src/polyswarmclient/transaction.py
```diff
@@ -190,10 +201,11 @@ def __init__(self, client, verifiers):
             if nonce_error:
                 logger.error('Nonce desync detected, resyncing nonce and retrying')
                 nonce_manager.mark_update_nonce()
-                if not has_required:
-                    continue
 
-            return True, {**result, **tx_result}
+            if retry:
```
Rizato: So the original point of the `has_required_event` was to not trigger a retry in the event of a successful later transaction, but not first. 

For instance, if someone sent an approve outside polyswarm client. So the next assertion approve would fail due to nonce, but the assertion would still succeed. We don't want to lose the successful assertion by retrying.

Rizato: I will approve for now while I write a POC test to trigger that.

mjkoo: Thanks, decided to wake up early and take a fresh stab at this probably gonna refactor it a bit

Rizato:
Couple nits regarding `sender already asserted` failures being retried, but it doesn't impact handling.

PR #197, opened 2019-02-18
Handle receipt timeouts (nonce too high due to transmission issues or…

mjkoo:
… other)


supernothing:
There a reasonable way to test the error this is fixing locally?

mjkoo:
Apply this patch and run orchestration
```
diff --git a/src/polyswarmclient/transaction.py b/src/polyswarmclient/transaction.py
index 31458ac..f927168 100644
--- a/src/polyswarmclient/transaction.py
+++ b/src/polyswarmclient/transaction.py
@@ -67,7 +67,7 @@ class NonceManager:
         """
         async with self:
             results = range(self.base_nonce, self.base_nonce + amount)
-            self.base_nonce += amount
+            self.base_nonce += amount + 1

             # Inside nonce_lock so that there is no way the next acquire could miss an in progress
             async with self.in_progress_condition:
@@ -207,7 +207,7 @@ class AbstractTransaction(metaclass=ABCMeta):

                 if self.client.tx_error_fatal:
                     logger.error('Received fatal transaction error during post', extra={'extra': results})
-                    exit(1)
+                    #exit(1)
                 elif not all_known_tx_errors:
                     logger.error('Received transaction error during post', extra={'extra': results})

@@ -274,7 +274,7 @@ class AbstractTransaction(metaclass=ABCMeta):
             if not success:
                 if self.client.tx_error_fatal:
                     logger.error('Received fatal transaction error during get', extra={'extra': results})
-                    exit(1)
+                    #exit(1)
                 else:
                     logger.error('Received transaction error during get', extra={'extra': results})
```

PR #199, opened 2019-02-20
Ensure results is never None coming from transaction send

mjkoo:


src/polyswarmclient/transaction.py
```diff
@@ -125,18 +125,21 @@ def __init__(self, client, verifiers):
...
-        transactions = result.get('transactions', [])
+        success, results = await self.client.make_request('POST',
+                                                          self.get_path(),
+                                                          chain,
+                                                          json=self.get_body(),
+                                                          send_nonce=True,
+                                                          api_key=api_key,
+                                                          tries=tries)
+
+        results = {} if results is None else results
```
mjkoo: This is the relevant fix

PR #205, opened 2019-02-24
We shouldn't error after cancelling tasks from handle_run

zv:
On Friday night I tracked this bug down to [`asyncio_stop@__init__.py#46`](https://github.com/polyswarm/polyswarm-client/blob/master/src/balancemanager/__init__.py#L46). 

After `handle_run`'s await is finished, the event loop concurrently proceeds to the `websocket` tasks in the call to `asyncio_stop`. I'm not familiar with all of the callers of `asyncio_stop` & `run.register(fn)` so I tried to make the solution to this problem as general as I thought it could require. 

With that said, I suspect there is a cleaner way to handle this issue from within callers of `asyncio_stop` and would like your opinion on another, possibly more direct way to handle this issue.


mjkoo:
I agree there is a cleaner way to handle this, as this primarily affects the one-shot deposit/withdraw runners theres no reason to even start up the websocket event loop.

have you looked into why this fails CI?

mjkoo:
https://github.com/polyswarm/polyswarm-client/tree/fix/cancelation-error-2 thoughts on this

mrtizmoatwork:
Closing in favor of #212

PR #206, opened 2019-02-25
Handle rate limits by short sleep between retries on a 429

mjkoo:
Tested with https://github.com/polyswarm/polyswarmd/tree/feature/simulate-rate-limits

Handles rate limits by inserting a short (2s) sleep between retries on encountering a 429, if we need to temporarily stop all tasks from requesting in order to recover from exceeding rate limits we'll need to add in some additional synchronization, however my feeling is that if that's the case we're gonna be dropping bounties to the floor and should instead re-evaluate our default rate limit settings (@supernothing any thoughts?)


supernothing:
I think what might help make this situation better is if we have a per-endpoint rate limit rather than a general one for all requests. Some of these requests (like posting bounties) are much heavier than ones just getting info, so putting them all in the same bucket seems silly if the idea is to mitigate DoS. As it is, I don't really have a problem raising the limits for now, since they don't do much to begin with, but we should work on getting proper limits in place.

For this pull, does `post_artifacts` need a retry as well? Any other endpoints?

mjkoo:
Agree re: per endpoint but yeah current way we're implementing doesnt allow this granularity.

Yeah re: post_artifacts, think thats it will verify

PR #223, opened 2019-03-17
Add callback for validating an API key recieved by click

zv:


src/polyswarmclient/config.py
```diff
@@ -6,6 +7,26 @@
...
+    try:
+        # If we recieved the parameter's default value, then don't bother
+        # checking.
+        if value == param.get_default(ctx):
+            return value
+        # int(x, 16) doesn't check newlines, etc. This ensures we've got a clean API key
+        if len(value) != 32:
+            raise click.BadParameter('API_KEY must be 32 characters')
+        if format(int(value, 16), 'x') != value:
+            raise click.BadParameter("API key is an invalid 32-byte hex value.")
```
mjkoo: 16 byte

mjkoo: 16 byte

src/polyswarmclient/config.py
```diff
@@ -6,6 +7,26 @@
...
+def validate_apikey(ctx, param, value):
+    """Validates the API key passed in through click parameters"""
+    try:
+        # If we recieved the parameter's default value, then don't bother
+        # checking.
+        if value == param.get_default(ctx):
+            return value
+        # int(x, 16) doesn't check newlines, etc. This ensures we've got a clean API key
+        if len(value) != 32:
+            raise click.BadParameter('API_KEY must be 32 characters')
```
mjkoo: Would prefer "API_KEY" and "API key" below to be consistent

src/polyswarmclient/config.py
```diff
@@ -6,6 +7,26 @@
...
+        if format(int(value, 16), 'x') != value:
```
mjkoo: Think id weakly prefer something like `if len(value) != 32 and not all([c in lower(string.hexdigits) for c in s])` or similar, but this is fine too

src/polyswarmclient/config.py
zv: lol yep, thanks.

src/polyswarmclient/config.py
zv: it is kindof confusing, i agree. resolved in c5112ee3ed3485635d2b7b61b88a8c441a4aaa85

src/polyswarmclient/config.py
zv: resolved a5c81e42c083c4d6e150deabd8c3da6a3894dc63

PR #227, opened 2019-03-22
[WIP] Update Docker base image to python:3.6-jessie

zv:
We should test the engines with an updated build here prior to merging this into stage and production.


mb2con:
We need to move off the jessie release.  https://superuser.com/questions/1417617/apt-get-update-is-failing-in-debian/1417633

PR #231, opened 2019-03-29
[WIP][Desires review]: Add lint tools

zv:
This change includes some lint configuration file that matches "polyswarm python style" as I see it but I'd still like everyone's feedback on how we can improve it.

It does contain both a `flake8` and `pylint` configuration, only because editor support is more spotty for `pylint`.

I haven't added any of these linters to code climate.

This does contain some changes recommended by the linter but only those that I thought were pretty obvious wins.


src/balancemanager/__init__.py
```diff
@@ -10,7 +10,7 @@
 RELAY_LEEWAY = 5
 
 
-class BalanceManager(object):
+class BalanceManager():
```
mrtizmoatwork: for class def'ns that do not inherit from another class, remove the (), so this would be `class BalanceManager:`. Same idea applies to the rest.

mrtizmoatwork:
@zv either finish this or close it.

zv:
closing

PR #246, opened 2019-04-12
Fix/sane backoff keyerror fix

aarongooch:


src/worker/__init__.py
```diff
@@ -12,7 +12,7 @@
 logger = logging.getLogger(__name__)
 
 REQUEST_TIMEOUT = 5.0
-
+BACKOFF_MAX = 30
```
Rizato: Is this read somewhere?

PR #250, opened 2019-04-24
Feature/multi task worker

Rizato:
In-worker scaling to go with infrastructure scaling later.


src/worker/__init__.py
```diff
@@ -65,7 +76,7 @@ def run(self):
                 try:
                     _, job = await redis.blpop(self.queue)
                     job = json.loads(job.decode('utf-8'))
-                    logger.info('Got job: %s', job)
+                    logger.info('Got job on task %s: %s', task_index, job)
```
mrtizmoatwork: let's standardize on string format like `"string {}".format(myvar)`

src/worker/__init__.py
```diff
@@ -111,7 +123,7 @@ def run(self):
                     'metadata': result.metadata,
                 })
 
-                logger.info('Scan results: %s', j)
+                logger.info('Scan results on task %s: %s', task_index, j)
```
mrtizmoatwork: let's standardize on string format like `"string {}".format(myvar)`

src/worker/__main__.py
```diff
@@ -83,7 +93,9 @@ def main(log, redis_addr, queue, api_key, backend, testing, log_format):
     scanner = scanner_class()
     init_logging(['worker', 'microengine', logger_name], log_format, loglevel)
 
-    worker = Worker(redis_addr, queue, api_key, testing, scanner)
+    logger.info('Running worker with %s tasks', tasks)
```
mrtizmoatwork: let's standardize on string format like `"string {}".format(myvar)`

src/worker/__main__.py
```diff
@@ -57,18 +58,27 @@ def choose_backend(backend):
               help='API key to use with polyswarmd')
 @click.option('--backend', envvar='BACKEND', required=True,
               help='Backend to use')
+@click.option('--tasks', envvar='TASKS', default=multiprocessing.cpu_count(),
+              help='Number of simultaneous worker tasks. (A task is read redis, download file and scan)')
```
mrtizmoatwork: oxford comma to make this more clear -> "A task is: read redis, download file, and scan"

Rizato:
@mrtizmoatwork Using `%s` in our logger statements and `{}.format()` elsewhere is more consistent with the rest of our code. 

@zv brought up f strings since we moved to 3.6 `f"{var_name}"`
I think that is probably the cleanest option. Do we want to start standardizing on those?

mrtizmoatwork:
@zv @Rizato yes. I'd like to go to f strings. I forgot that python 3.6 was the version where it was added. I'd prefer that we use the same string formatting system for all strings, not just logger statements though.

PR #252, opened 2019-04-25
Remove ethereum module

zv:
This patch removes all instances of the `ethereum` module referenced in `polyswarm-client`'s Python code, ending a slew of Windows build issues arising from this package's dependencies.

Today, the amount of code referencing the `ethereum` module in `polyswarm-client` has shrunk to the extent that only it's implementation of sha3/Keccak remains; All usages of this `sha3` function have been replaced with the reference implementation upstream along with unit tests to verify absence of regressions in functions calling `ethereum.utils.sha3`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ base58==0.2.5
 git+https://github.com/polyswarm/python-clamd.git@async#egg=clamd
 click==6.7
 coverage==4.5.1
-ethereum==2.3.2
+pycryptodome>=3.4.7
 hypothesis==3.82.1
 pytest==3.9.2
 pytest-asyncio==0.9.0
--- a/docker/windows/base/polyswarm-profile.ps1
+++ b/docker/windows/base/polyswarm-profile.ps1
@@ -57,8 +57,8 @@ function Install-Service {
     #>
 
     Param(
-        [Parameter(Mandatory = $true)][String] $Name,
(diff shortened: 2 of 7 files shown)


requirements.txt
```diff
@@ -8,7 +8,7 @@ base58==0.2.5
 git+https://github.com/polyswarm/python-clamd.git@async#egg=clamd
 click==6.7
 coverage==4.5.1
-ethereum==2.3.2
+pycryptodome>=3.4.7
```
mrtizmoatwork: @zv this is great work, but if we are going to change it, let's go with the actual python hashlib: https://docs.python.org/3.6/library/hashlib.html

zv:
```
>>> from Crypto.Hash import keccak
>>> import hashlib
>>> keccak.new(digest_bits=256, data=b"1").hexdigest() == hashlib.sha3_256(b"1").hexdigest()
False
```
-----------------------------------------------------------------------------------------------

I don't recall the details but there are minor parameter differences between Keccak implementation beyond the output size which affect digest. 

I once submitted an [implementation](https://github.com/zv/SHA3ImplementedInsideofaRARfile) to a the creators extended code implementation gallery and was classed with generic Keccak implementation rather than FIPS or some specific Keccak parameter family.

mrtizmoatwork:
ok. go for it

PR #260, opened 2019-05-03
Catch OsError and ReplyError on redis calls

Rizato:
In both producer and worker


src/microengine/producer.py
```diff
@@ -60,6 +60,10 @@ def __init__(self, client, testing=0, scanner=None, chains=None):
                 j = json.loads(result[1].decode('utf-8'))
                 return j['index'], ScanResult(bit=j['bit'], verdict=j['verdict'], confidence=j['confidence'],
                                               metadata=j['metadata'])
+            except aioredis.errors.ReplyError:
+                logger.critical('Redis out of memory')
```
aarongooch: use `logger.exception` to output the stacktrace.

src/microengine/producer.py
```diff
@@ -60,6 +60,10 @@ def __init__(self, client, testing=0, scanner=None, chains=None):
...
+            except OSError:
+                logger.error('Redis connection down')
```
aarongooch: same here `logger.exception`

src/microengine/producer.py
```diff
@@ -72,13 +76,18 @@ def __init__(self, client, testing=0, scanner=None, chains=None):
...
 
-        # Age off old result keys
-        await self.redis.expire(key, KEY_TIMEOUT)
+            # Age off old result keys
+            await self.redis.expire(key, KEY_TIMEOUT)
 
-        return [results.get(i, ScanResult()) for i in range(num_artifacts)]
+            return [results.get(i, ScanResult()) for i in range(num_artifacts)]
+        except OSError:
+            logger.error('Redis connection down')
```
aarongooch: and here `logger.exception`

src/worker/__init__.py
```diff
@@ -88,6 +88,10 @@ def run(self):
 
                     index = job['index']
                     chain = job['chain']
+                except OSError:
+                    logger.error('Redis connection down')
```
aarongooch: and here

src/worker/__init__.py
```diff
@@ -127,4 +131,9 @@ def run(self):
                 logger.info(f'Scan results on task {task_index}', extra={'extra': j})
 
                 key = f'{self.queue}_{guid}_{chain}_results'
-                await redis.rpush(key, j)
+                try:
+                    await redis.rpush(key, j)
+                except OSError:
+                    logger.error('Redis connection down')
```
aarongooch: and here

PR #266, opened 2019-05-06
Build, test and install from PSC root in Windows.

zv:
This copies, builds and installs `polyswarm-client` directly, no longer pulling from pypi after release, removing any historical cruft not needed as well.

It also adds a `container_test` step which installs & runs `tox` inside a Windows container.


mrtizmoatwork:
I'll need to think on this over night. My initial thought is that we should have 2 separate jobs in there. 1) build from the code in the repo, use it for windows tests, use it for the e2e pipeline 2) build from pypi, use it for prod engines... or something like that.

PR #269, opened 2019-05-09
Add --denomination and --all options to balancemanager

Rizato:
Allow users to select `nct`, `nct-gwei`, or `nct-wei` when entering amount

For withdrawals and deposits there is an `--all` option for transferring the entirety of one's nct balance.

~Still need to add denominations to maintainer~


src/balancemanager/__init__.py
```diff
@@ -161,8 +188,8 @@ def run(self):
         home_balance = await self.client.balances.get_nct_balance(chain='home')
         side_balance = await self.client.balances.get_nct_balance(chain='side')
         if home_balance >= self.refill_amount:
-            logger.info('Sidechain balance (%s NCT) is below minimum (%s NCT). Depositing %s NCT', side_balance,
-                        self.minimum, self.refill_amount)
+            logger.info(f'Sidechain balance of {side_balance} nct-wei is under minimum {self.minimum} nct-wei.'
```
mrtizmoatwork: need a space at the end of the first string, so the 2 sentences are separated by a space.

PR #270, opened 2019-05-10
Add artifact type handling

Rizato:


src/polyswarmclient/abstractambassador.py
```diff
@@ -83,16 +85,18 @@ def connect(cls, polyswarmd_addr, keyfile, password, api_key=None, testing=0, in
         """
         pass
 
-    async def push_bounty(self, amount, ipfs_uri, duration, chain, api_key=None):
+    async def push_bounty(self, amount, ipfs_uri, duration, chain, api_key=None, artifact_type=ArtifactType.FILE):
         """Push a bounty onto the queue for submission
 
         Args:
+            artifact_type (str): Type of artifact being pushed
```
mrtizmoatwork: should be in the same order as the func params ^^

PR #272, opened 2019-05-10
Run kick_downstream.py after all other tasks

zv:


.gitlab-ci.yml
```diff
@@ -91,5 +91,13 @@ linuxbase:
     - tags
   script:
     - docker login -u "$CI_CUSTOM_DOCKER_HUB_USERNAME" -p "$CI_CUSTOM_DOCKER_HUB_PASSWORD"
-    - docker build -t polyswarm/polyswarm-client -f docker/Dockerfile .
-    - docker push polyswarm/polyswarm-client
+
+kickafterpacker:
+  before_script:
```
mrtizmoatwork: remove this before_script, instead define an image key: use the image from our internal docker registry (stage image). It already has these things installed.

.gitlab-ci.yml
```diff
@@ -91,5 +91,13 @@ linuxbase:
     - tags
   script:
     - docker login -u "$CI_CUSTOM_DOCKER_HUB_USERNAME" -p "$CI_CUSTOM_DOCKER_HUB_PASSWORD"
-    - docker build -t polyswarm/polyswarm-client -f docker/Dockerfile .
-    - docker push polyswarm/polyswarm-client
+
+kickafterpacker:
```
mrtizmoatwork: change this name to something more specific like "kickpscw"

.gitlab-ci.yml
```diff
@@ -91,5 +91,13 @@ linuxbase:
...
+  before_script:
+    - apk --no-cache add py3-requests
+  stage: notifydependents
+  only:
+    - feature/windows-builds
+    - master
```
mrtizmoatwork: this will be used to kick polyswarm-client-windows and all linux engines, so I also add "feature/linux-builds"

.gitlab-ci.yml
```diff
@@ -91,5 +91,13 @@ linuxbase:
     - tags
   script:
     - docker login -u "$CI_CUSTOM_DOCKER_HUB_USERNAME" -p "$CI_CUSTOM_DOCKER_HUB_PASSWORD"
-    - docker build -t polyswarm/polyswarm-client -f docker/Dockerfile .
-    - docker push polyswarm/polyswarm-client
```
mrtizmoatwork: why are you deleting these 2 lines? ^ Without them, that job is pointless.

zv: That's absolutely right, that was accidental.

PR #273, opened 2019-05-10
Standardize exit on failures and improve failure logging

myhndl:
…f insufficient balance message on arbiter.

I believe warrants a psc version bump. Wanna check for any side-effects first.


mrtizmoatwork:
@Rizato are you ok with this one?

PR #282, opened 2019-05-13
Remove vesitigal verbatimdb

myhndl:


myhndl:
Pretty sure I misunderstood verbatimdb on this one, had assumed that the verbatimdb folder was vestigial at this point. Seems like it's not, closing PR.

PR #296, opened 2019-05-23
provide a setup() method for the Scanner() class to use to perform se…

mrtizmoatwork:
…tup tasks before doing scanning.

### DIFF
--- a/src/worker/__init__.py
+++ b/src/worker/__init__.py
@@ -72,6 +72,11 @@ class Worker(object):
     async def setup(self):
         self.scan_lock = asyncio.Semaphore(value=self.scan_limit)
         self.download_lock = asyncio.Semaphore(value=self.download_limit)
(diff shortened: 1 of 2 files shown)


src/worker/__init__.py
```diff
@@ -72,6 +72,11 @@ def run(self):
     async def setup(self):
         self.scan_lock = asyncio.Semaphore(value=self.scan_limit)
         self.download_lock = asyncio.Semaphore(value=self.download_limit)
+        if not await self.scanner.setup():
+            logger.critical('Scanner instance reported unsuccessful setup. Exiting.')
+            asyncio_stop()
+            asyncio_join()
+            exit(0)
```
Rizato: We probably want a non-zero exit here, right?

mrtizmoatwork: yup. set to 1.

PR #301, opened 2019-06-05
Use polyswarm-client Verdict metadata schema in default Scanners

Rizato:


src/microengine/multi.py
```diff
@@ -29,18 +30,28 @@ def __init__(self):
...
         bits = [r.bit for r in results]
         verdicts = [r.verdict for r in results]
         confidences = [r.confidence for r in results]
-        metadatas = [r.metadata for r in results]
+        try:
+            metadatas = [json.loads(r.metadata) for r in results
+                         if r.metadata and Verdict.validate(json.loads(r.metadata))]
+            # Only combining the family name for now.
+            combined = Verdict()\
+                .set_malware_family(','.join([metadata.get('malware_family', '') for metadata in metadatas])).json()
```
mrtizmoatwork: if more than one engine return unique malware family, this is going to make a mess in the portal UI. First, let's make this a set() of strings, so each is unique. Second, let's assume that the engine author has some logic right here that smartly combines the malware family info from the multiple backends into a single result. So, no join, no list, just one malware family. So for the same of this exemplar muli-scanner, just put them into a set, get the first non-empty value and use it.

mrtizmoatwork: Also, add a comment above this line that says something like: "# the author is responsible for distilling multiple backend metadata content into a single result to include in the ScanResult()"

src/microengine/yara.py
```diff
@@ -27,10 +28,16 @@ def __init__(self):
             ScanResult: Result of this scan
         """
         matches = self.rules.match(data=content)
+        sysname, _, _, _, machine = os.uname()
+        metadata = Verdict().set_scanner(operating_system=sysname,
+                                         architecure=machine,
+                                         vendor_version=yara.__version__)
         if matches:
-            return ScanResult(bit=True, verdict=True)
+            metadata.set_malware_family(','.join([match.rule for match in matches]))
```
mrtizmoatwork: no join() here. They only get to include a single value. So, same as the engine-multi example, put a comment above this line that says: "# the author is responsible for distilling multiple metadata content values into a single result to include in the ScanResult()"

So, for this example, put the results into a set, and grab the first non-empty value.

PR #303, opened 2019-06-06
Fix/ambassador bounty envvars

aarongooch:


HISTORY.md
```diff
@@ -1,5 +1,9 @@
 # Release History
 
+### 2.0.3 (2019-06-06)
+
+* **Feature** - Add Ambassador bounty submission env vars to dynamically set rate of bounty submissions.
+
```
Rizato: Add a fix element for my latest squash and merge into master

Rizato: `Pass artifact type on second on_bounty_post_failed call`

Rizato: `Remove stop and join when scanner setup fails`

PR #310, opened 2019-06-11
Add BidStrategyBase class that holds min and max multipliers for assertion bids

Rizato:
Create BidStrategyBase
Read multipliers from base given to Microengine. 
Multiply contract minimum amount by the multipliers to get the bounty range, before the confidence is taken into account.

Still needs to support handling modifications based on filetype.


Rizato:
Removed WIP for now. The file type bid modifications will come later when ambassador provides metadata on bounties.

PR #314, opened 2019-06-12
Refactoring ci

sbnetops:
Refactoring CI to have stages as:

Build -> Test -> Release -> Deploy

And jobs inside each stage run in parallel. This is built on top of the https://12factor.net/ philosophy.

- Master branch triggers releases for latest docker images
- Tags trigger releases for tagged docker images and pypi releases
- Build and Test stages run on every push
- Deploy runs on master branch
- Filter pypi releases based on tags in the format X.X.X, do execute for other tags

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,40 +4,112 @@ services:
   - docker:dind
 
 stages:
+  - build
   - test
   - e2e
-  - test_pypi
-  - pypi
-  - container_test
-  - postrelease
-  - notifydependents
+  - release
+  - deployment
 
 variables:
   STAGE_IMAGE: $REPO_URL/stage:latest
   DOCKER_HOST: tcp://localhost:2375/
 
-test:
-  image: $STAGE_IMAGE
+###############################################################
+# Build Stage (jobs inside a stage run in parallel)
+###############################################################
+
+build-linux:
+  stage: build
+  tags:
+    - kube
+  script:
+    - docker pull "$REPO_URL/polyswarm-client:latest"
+    - docker build -t polyswarm/polyswarm-client -f docker/Dockerfile --cache-from="$REPO_URL/polyswarm-client:latest" .
+    - docker push "$REPO_URL/polyswarm-client:$CI_COMMIT_SHA"
+
+
+build-windows:
+  stage: build
+  tags:
+    - win-participants
+  script:
+    - aws ecr get-login --no-include-email --region us-east-1 | Powershell
+    - docker pull "$REPO_URL/polyswarm-client-windows:latest"
+    - docker build -t polyswarm/polyswarm-client-windows -f docker/windows/Dockerfile --cache-from="$REPO_URL/polyswarm-client-windows:latest" .
+    - docker push "$REPO_URL/polyswarm-client-windows:$CI_COMMIT_SHA"
+
+###############################################################
+# Test Stage
+###############################################################
+
+test-linux:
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -4,41 +4,113 @@ services:
...
+  stage: test
   tags:
     - kube
-  only:
-    - feature/test-pypi
   script:
+    # TODO: Maybe consider executing this inside the polyswarm-client image we just built
     - python setup.py sdist bdist_wheel
     # using env variables from Gitlab: TWINE_USERNAME, TWINE_PASSWORD, TWINE_REPOSITORY_URL
     - twine upload --repository-url $TWINE_REPOSITORY_URL_TEST -u $TWINE_USERNAME_TEST -p $TWINE_PASSWORD_TEST dist/*
```
mrtizmoatwork: be careful putting this in the path. If there is a successful push to the test pypi of a version of the package. and you try to re-run this job, all subsequent executions of this job will fail. because you can only push a specific version to pypi one time.

.gitlab-ci.yml
```diff
@@ -4,41 +4,113 @@ services:
...
 
 variables:
   STAGE_IMAGE: $REPO_URL/stage:latest
   DOCKER_HOST: tcp://localhost:2375/
 
-test:
-  image: $STAGE_IMAGE
+###############################################################
+# Build Stage (jobs inside a stage run in parallel)
+###############################################################
```
mrtizmoatwork: we are using polyswarm-client from the source or from pypi to build these images?

sbnetops: We are using `polyswarm-client` from the source to build the images.

.gitlab-ci.yml
sbnetops: I was talking to @Rizato about it, we will probably only run pypi tests and releases on tagged executions. We still have to figure out what will be the tag pattern that will trigger pypi executions tough.

sbnetops:
I will also take a look at the `Dockerfile`s :+1:

.gitlab-ci.yml
```diff
@@ -4,45 +4,127 @@ services:
...
 
 test-push-to-pypi:
-  image: $STAGE_IMAGE
-  stage: test_pypi
+  stage: test
   tags:
     - kube
   only:
-    - feature/test-pypi
+    - /^[0-9]+\.[0-9]+\.[0-9]+rc[0-9]+$/
```
mrtizmoatwork: this regex is not necessary. you can just do "tags" and it will apply to all tags.

.gitlab-ci.yml
```diff
@@ -4,45 +4,127 @@ services:
...
+  except:
```
mrtizmoatwork: if you do "only: tags", then you do not need this "except" part.

.gitlab-ci.yml
```diff
@@ -4,45 +4,127 @@ services:
...
+    # Pushes to DockerHub
+    - docker logout
+    - docker login -u "$CI_CUSTOM_DOCKER_HUB_USERNAME" -p "$CI_CUSTOM_DOCKER_HUB_PASSWORD"
+    - docker push "polyswarm/polyswarm-client:latest"
+
+push-to-pypi:
+  stage: release
+  tags:
+    - kube
+  only:
```
mrtizmoatwork: should be "only: tags" and then remove the "except" part.

PR #326, opened 2019-06-17
Add signal.SIGINT handler on windows

Rizato:


Rizato:
Per https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/signal_windows.go

PR #332, opened 2019-07-03
Bounty metadata changes with windows support

Rizato:


tests/test_bounty_filter.py
```diff
@@ -0,0 +1,124 @@
...
+    # act
+    padded = BountyFilter.pad_metadata(metadata, 2)
+    # assert
+    assert padded == [{"mimetype": "text/plain"}, None]
+
+
+@pytest.mark.asyncio
+async def test_pad_fills_with_none_on_invalid_metadata():
+    # arrange
+    metadata = [{"asdf": "asdf"}]
```
mrtizmoatwork: add another test checking {"mimetype": "bogus"} would be good to see. Do we check anywhere to see whether the value is a real/valid mimetype?

Rizato: I don't check anywhere. I could add something to make sure it is at least `something/something`

mrtizmoatwork: let's just leave it as-is for now.

PR #338, opened 2019-07-10
remove tox.ini, tox from requirements, and change test stage to use p…

mb2con:
remove tox.ini, tox from requirements.txt, and change test stage to use pytest instead of tox

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -64,7 +64,7 @@ test-linux:
     - kube
   script:
     - docker pull $REPO_URL/$BASE_IMAGE_NAME:${MANUAL_IMAGE_TAG:-$CI_COMMIT_SHA}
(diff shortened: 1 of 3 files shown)


mrtizmoatwork:
We determined the following:
1. Right now, we really only need to be testing for python 3.6.5. But, as python 3.7 and 3.8 become more and more common, we may want to add testing for those versions as well.
2. If we remove tox from the polyswarm-client docker image, than for the microengines that are using it, we will need to install tox in their images before doing testing.

I would prefer that we do not have testing tools in our production polyswarm-client image. So maybe we should create an image with a `:test` tag and that will be the one that has the testing tools installed.

mrtizmoatwork:
I'm closing this. We can re-evaluate and revise our testing methods as part of a bigger project.

PR #340, opened 2019-07-26
Indicate metadata should be json-parsable string

CalebFenton:
It looks like metadata is intended to be parsable as a json string and this should be reflected in the comments.


Rizato:
Thanks @CalebFenton

PR #341, opened 2019-07-27
[WIP-PROPOSAL] Add a signal handler if `add_signal_handler` is unimplemented.

zv:
This PR contains a proposal to change the exception handler for `NotImplemented` raised from `loop.add_signal_handler` to stop throwing an exception and instead install a traditional signal handler.


Rizato:
Does this work? When I first set it up the base signal handler was not working cleanly with asyncio.

PR #344, opened 2019-08-01
Use JQ syntax for selecting the key/value in metadata

Rizato:
Didn't mean to merge the earlier PR, but here is the requested feature to use JQ syntax instead of a plain key for creating filters.


Rizato:
Well, the jq library does not support windows.

PR #347, opened 2019-08-12
Do not discard results when running solo microengines

sbnetops:
When running microengines that are not in the producer/consumer model, the original `fetch_and_scan()` code discards the `ScanResult()` instance from the scanner, returning an empty `ScanResult()` instead, which means "no assertion". This fix uses the existing `ScanResult()` in case the bit mask is set.

This bug was compromising our `e2e` tests. Thanks @mrtizmoatwork for the help looking into it.


mrtizmoatwork:
@Rizato was this an oversight? Or was there a reason this return value was omitted?

PR #350, opened 2019-08-16
potential fix for the hung engine issue

supernothing:
One of my previous PRs, #330 , introduced a bug that can cause engines to permanently block under load. A `blpop` timeout of 0 indicates blocking indefinitely, not non-blocking. It essentially reduced to the previous problematic behavior.

Replaced with a normal `lpop`, which is atomic and should work as this was intended to. From testing, engines now seem to recover properly after overloading events.


src/polyswarmclient/producer.py
```diff
@@ -106,7 +106,7 @@ def __init__(self, client, redis_uri, queue, time_to_post, bounty_filter=None, c
                 results = {r[0]: r[1] for r in results if r is not None}
 
                 # Age off old result keys
-                await self.redis.expire(key, KEY_TIMEOUT)
+                await self.redis.expire(key, KEY_TIMEOUT+10)
```
Rizato: Should we update key timeout for this?

supernothing: The issue I was trying to fix was that this races the `wait_for_result` function and sometimes may expire the key too early (which I think might have contributed to those `blpop`s blocking). I can create a separate variable for it.

PR #352, opened 2019-08-22
Add bid per artifact

Rizato:
Support the bid per artifact changes in polyswarmd and contractor. 

TODO
- [x] Update tests
- [x] Modify bid so that bid % sum(bid_portions) is always 0, and no extra 1/10^18ths NCT are given to the arbiters, regardless of verdicts


src/ambassador/filesystem.py
```diff
@@ -55,20 +55,28 @@ def __init__(self, client, testing=0, chains=None, watchdog=0, submission_rate=3
 
         while True:
             try:
-                filename = random.choice(self.artifacts)
+                num_artifacts = min(255, len(self.artifacts))
```
mrtizmoatwork: is it supposed to be 255 or 256?

src/polyswarmclient/bountiesclient.py
```diff
@@ -328,7 +328,7 @@ def __init__(self, client):
         Args:
             bounty_guid (str): The bounty which we have asserted on
             index (int): The index of the assertion to reveal
-            nonce (str): Secret nonce used to reveal assertion
+            nonce (str): Secret nonce used to reveal assertio
```
mrtizmoatwork: assertion

src/ambassador/filesystem.py
Rizato: 255, but it should also be len -1. 
num_artifacts is also a bad name, there.

Rizato: Should be 256. For some reason I was thinking the last value was inclusive. (But the len would still be wrong at that point)

src/polyswarmclient/__init__.py
```diff
@@ -360,6 +360,15 @@ def to_wei(amount, unit='ether'):
     def from_wei(amount, unit='ether'):
         return w3.fromWei(amount, unit)
 
+    @staticmethod
+    def get_artifact_bid_at_(mask, bid, index):
+        if not mask[index]:
+            return 0
+
+        bid_index = len([m for m in mask[:index] if m])
+        logger.critical(mask[:index])
```
sbnetops: Is there a reason to have a critical log here? Everything seems to keep working after this.

src/polyswarmclient/abstractmicroengine.py
```diff
@@ -16,6 +16,10 @@
 logger = logging.getLogger(__name__)
 
 
+class InvalidBidError(Exception):
```
sbnetops: Would be great to have a base exception such as `PolyswarmClientException` that all exceptions inherit from. Also, putting all exceptions together in one module would help developers check what kind of error they can expect.

src/polyswarmclient/__init__.py
Rizato: Forgot about it. 
Also, need to change the bid_index logic there.

src/polyswarmclient/abstractmicroengine.py
```diff
@@ -103,14 +104,15 @@ def connect(cls, polyswarmd_addr, keyfile, password, api_key=None, testing=0, in
             chain (str): Chain we are operating on
 
         Returns:
-            int: Amount of NCT to bid in base NCT units (10 ^ -18)
+            Bid: Amount of NCT to bid in base NCT units (10 ^ -18)
```
supernothing: This should be `list[int]`, shouldn't it?

src/polyswarmclient/filters/bountyfilter.py
```diff
@@ -67,14 +67,14 @@ def is_allowed(self, metadata):
...
-                                                                                     "accept": self.accept}})
+            logger.debug('Metadata not accepted. Skipping artifact', extra={"extra": {"metadata": metadata,
+                                                                            "accept": self.accept}})
             return False
 
         rejected = any([f.filter(metadata) for f in self.reject])
         if self.reject and rejected:
-            logger.info('Metadata rejected. Skipping artifact', extra={"extra": {"metadata": metadata,
-                                                                                 "reject": self.reject}})
+            logger.debug('Metadata rejected. Skipping artifact', extra={"extra": {"metadata": metadata,
```
supernothing: I'm gonna channel some @mrtizmoatwork here and say "single quotes"

src/polyswarmclient/filters/bountyfilter.py
```diff
@@ -67,14 +67,14 @@ def is_allowed(self, metadata):
         accepted = any([f.filter(metadata) for f in self.accept])
 
         if self.accept and not accepted:
-            logger.info('Metadata not accepted. Skipping artifact', extra={"extra": {"metadata": metadata,
-                                                                                     "accept": self.accept}})
+            logger.debug('Metadata not accepted. Skipping artifact', extra={"extra": {"metadata": metadata,
```
supernothing: quotes

Rizato:
@supernothing Can you take another look?

PR #357, opened 2019-08-26
Remove platform assumptions where it makes sense

myhndl:


Rizato:
The eicar fix got pushed into master some time back.

PR #359, opened 2019-09-06
Decode content according to artifact type before passing to scan

Rizato:
Uses polyswarm-artifact to decode the content before scanning


Rizato:
This won't work until the polyswarm-artifact PR is merged & a tag pushed

Rizato:
@mrtizmoatwork Can I get another review on this?

PR #360, opened 2019-09-10
Bump from 2.4.0 to 2.4.1

Rizato:


Rizato:
Thanks for the title note.

PR #364, opened 2019-09-12
Add tool to check liveliness of running particpants

Rizato:
Adds tools to gather liveliness data, and check it using files. 

Checks liveliness by recording the time since the last loop update, and the average wait time for bounties in progress. 

For ambassador, microengine, and arbiter wait times are in blocks, so that a down polyswarmd won't cause liveliness checks to fail. 
Worker does time in seconds since it doesn't depend on the polyswarmd websocket.


src/polyswarmclient/liveliness.py
```diff
@@ -0,0 +1,225 @@
...
+                with FileLock(f.fileno()):
+                    content = f.read()
+                    logger.debug('Liveliness contents %s', content)
+                    liveliness = Liveliness.from_json(content)
+            except OSError:
+                logger.exception('Unable to lock file')
+                return False
+
+        time_since_last_loop = int(time.time()) - liveliness.last_iteration
+        print('Last loop was {0} seconds ago, and the average wait time is {1}'.format(time_since_last_loop,
```
supernothing: What happens when there aren't bounties going through? Still reports live?

src/polyswarmclient/liveliness.py
```diff
@@ -0,0 +1,225 @@
...
+        self.guid = guid
+        self.start_time = start_time
+
+    def __eq__(self, other):
+        # Only care about guid for __eq__
+        return isinstance(other, self.__class__) and other.guid == self.guid
+
+    def __hash__(self):
+        # Only care about guid for __hash__
+        h = 11
```
supernothing: wut

Rizato: Can you be more specific? 
I implemented if because I implemented __eq__
This is the way I learned to implement hash functions for custom objects. Grab a couple prime numbers and do all the adding and multiplying.

src/polyswarmclient/liveliness.py
Rizato: This part is only a check that the event loop is not blocked. 

The other one detects the rate of bounties being solved, successful or not, to make sure we aren't blocking somewhere in the bounty handling, but in a way that doesn't block the loop itself.

src/polyswarmclient/liveliness.py
Rizato: Can you be more specific?
I implemented if because I implemented eq
This is the way I learned to implement hash functions for custom objects. Grab a couple prime numbers and do all the adding and multiplying.

PR #366, opened 2019-09-16
Use wait_for() on the coroutine instead of manual timeouts

Rizato:
The goal here is exit all of the wait_for_result coroutines near simultaneously, but not lose any valid responses provided before things timed out.


Rizato:
In local testing
The old way took 40.15 seconds to timeout, new way to 21.06

PR #370, opened 2019-10-02
Add environment controller artifacts per bounty in file system ambassador

Rizato:
256 artifacts per bounty causes timeouts and using that as the default in testing causes issues. 
This sets a lower default, but allows users to up that value if they want.


Rizato:
Code climate issues are about our use of random, but it's not a security critical random.

PR #372, opened 2019-10-09
Add an error level log when connection is reestablished

Rizato:
Used Error level to match the waiting n second logs. 
It no longer looks we things are hanging on a reconnect.


Rizato:
I'm not sure which line you mean @aarongooch

PR #373, opened 2019-10-10
Add ability to limit scans globally for producers

Rizato:
Add `AbstractRateLimit` class. 
Add `RedisDailyRateLimit` implementation. 
Honor `RedisDailyRateLimit` in `microengine::producer` and `arbiter::producer`. 

This opens the door to custom rate limits, but does not let users load a custom rate limit module, yet. 

Use case is if a user has a license for only N scans per period with a vendor.


Rizato:
Yea

PR #374, opened 2019-10-10
Use the default api endpoint as default polyswamrd-addr

Rizato:


myhndl:
A few changes:
1. default to marketplace polyswarmd addr only if `--testing` is not used
2. if `--testing` is used, default to `polyswarmd:31337`, not `localhost:31337`

The former prevents accidental marketplace connection. The latter is more accurate for what happens in local testing (Docker will spin up a daemon at `polyswarmd:31337`

myhndl:
Upon further discussion, scratch my above comment.

myhndl:
Assuming this works (I haven't tested yet), this PR looks good to me.

PR #375, opened 2019-10-11
Revert "Use the default api endpoint as default polyswamrd-addr"

Rizato:
Reverts polyswarm/polyswarm-client#374


Rizato:
Did it with the UI, but wasn't verified.

PR #376, opened 2019-10-11
Support paths in polyswarmd addresses

myhndl:


myhndl:
I tested this with a mock engine, was able to connect to both `polyswarmd:31337` (hosted via orchestration) and `api.polyswarm.network/v1/default`

src/polyswarmclient/__init__.py
```diff
@@ -234,6 +236,7 @@ def run(self, chains=None):
                     except (ValueError, aiohttp.ContentTypeError):
                         response = await raw_response.read() if raw_response else 'None'
                         logger.error('Received non-json response from polyswarmd: %s', response)
+                        logger.error('uri: %s', uri)
```
Rizato: Can you include this in the previous error log?

src/polyswarmclient/__init__.py
```diff
@@ -330,7 +333,9 @@ def sign_transactions(self, transactions):
         if not is_valid_ipfs_uri(ipfs_uri):
             raise ValueError('Invalid IPFS URI')
 
-        uri = urljoin(self.polyswarmd_uri, '/artifacts/{0}/{1}'.format(ipfs_uri, index))
+        uri = self.polyswarmd_uri + '/artifacts/{0}/{1}'.format(ipfs_uri, index)
```
Rizato: Can we convert this to f string
```suggestion
        uri = f'{self.polyswarmd_uri}/artifacts/{ipfs_uri}/{index}'
```
https://github.com/polyswarm/polyswarm-client/blob/fe4deb4a52862c6f1ab91d3784249a9e8cb52911/src/worker/__init__.py#L103

src/polyswarmclient/__init__.py
```diff
@@ -484,6 +491,7 @@ def get_artifacts(self, ipfs_uri, api_key=None):
                         except (ValueError, aiohttp.ContentTypeError):
                             response = await raw_response.read() if raw_response else 'None'
                             logger.error('Received non-json response from polyswarmd: %s', response)
+                            logger.error('uri: %s', uri)
```
Rizato: Combine error messages here

src/polyswarmclient/__init__.py
```diff
@@ -434,7 +439,9 @@ def get_artifacts(self, ipfs_uri, api_key=None):
             (str): IPFS URI of the uploaded artifact
         """
 
-        uri = urljoin(self.polyswarmd_uri, '/artifacts')
+        uri = self.polyswarmd_uri + '/artifacts'
```
Rizato: Can we do f-string here too

src/polyswarmclient/__init__.py
```diff
@@ -199,7 +200,8 @@ def run(self, chains=None):
         # Ensure we try at least once
         tries = max(tries, 1)
 
-        uri = urljoin(self.polyswarmd_uri, path)
+        uri = self.polyswarmd_uri + path
```
Rizato: f-string

src/polyswarmclient/__init__.py
myhndl: done

myhndl:
> I think like 300 could us an f-string too.

Done

src/polyswarmclient/__init__.py
myhndl: done

src/polyswarmclient/__init__.py
myhndl: done

src/polyswarmclient/__init__.py
myhndl: done

PR #378, opened 2019-10-11
remove newline from vendor_version

JavierBotella:


src/microengine/clamav.py
```diff
@@ -38,7 +38,7 @@ def __init__(self):
         vendor = await self.clamd.version()
         metadata = Verdict().set_scanner(operating_system=platform.system(),
                                          architecture=platform.machine(),
-                                         vendor_version=vendor)
+                                         vendor_version=vendor.strip('\n'))
```
myhndl: Would `.rstrip()` be more fault tolerant here? Might have different newlines on Windows?

JavierBotella: i just choice `.strip()` because i don't trust to much in the whole line. :) But could works well sure.

mrtizmoatwork: Yes. We need to use `.rstrip()` so it can handle all forms of trailing whitespace, not just `\n`.

JavierBotella: https://www.programiz.com/python-programming/methods/string/rstrip

what i am missing there? it's same of .strip but from the right since i know. You have to specify the things to remove.

mrtizmoatwork: oh. yeah, you are correct. sorry. This is fine.

PR #381, opened 2019-10-16
Prevent crash if liveliness is None

Rizato:
This stops crashing if liveliness is none, but it does no sort out why liveliness would be None at a random point while microengine is running. 

It should be None at the start or end, but I suspect the file lock is not working properly and somehow the liveliness check read a file during a write.


sbnetops:
Ok, I kind of get what is happening here. 

```python
    def write_sync(self, path, content):
        """ Write the given content to the file at the given path.

        Args:
            path: file path to write to
            content: content to write into the file
        """
        with open(path, 'w') as f:
            try:
                with FileLock(f.fileno()):
                    f.write(content)
            except OSError:
                logger.exception('Unable to lock file')
```
Here you are opening the file, truncating it and the acquiring the lock. If anyone acquires the lock before that, they will see an empty file. Also, if you release the lock and another file gets it before you run `__exit__()` and the buffer was not flushed, they will also see an empty file. Not to mention that `/tmp` might be mapped to memory and locks might behave differently there.

We should certainly **not** be using files for locks and avoid using local files for inter-process communication (specially when we can't control access to them appropriately, meaning that these are two processes that don't share a os-level lock, which is the case here). These are bad practices that will certainly generate issues. That said, here are some alternatives.

- One **possible** solution keeping the current pattern is the use of `lockfiles`. I can investigate a bit more into this.
- We could also create a unix-socket and write to it in order to communicate (just as the docker daemon does), but I believe that would only work on unix-based systems.
- Another possibility is to actually expose an endpoint and ask on that endpoint (even if we only bind to the localhost).
- Another possibility is to have a file that we just append at the end, therefore we never truncate the file. We would still need to write some kind of timestamp to the file and check if a reasonable amount of time has passed (or write cycles, more than one).
- Another possibility is to investigate how to have shared memory region. This will certainly require a lot of mambo jambo to figure out which one is the correct process we want to share memory with, get the os to let us read that area and we would certainly fall into the issue of having a critical region again, not to mention the differences between unix and windows.

In my opinion, the only real, sane way to do this is to expose an endpoint and talk to the process through it. It is plataform-independent and we will be certain that there are no race conditions of any kind. All other solutions will be questionable at best.

Let me know how you guys would like to proceed @mrtizmoatwork @Rizato

Rizato:
Shared memory was my first intent with this. However, there doesn't seem to be a way to do it on windows. There are windows calls to do it, but I couldn't find any good libs for it. 

We did discuss an endpoint, which would definitely be the cleanest solution. We don't want to be zoom though. I think we ruled it out early because of the security risks. I agree, it is the best option.

Rizato:
We could always make the liveliness endpoint a flag so people aren't surprised

mrtizmoatwork:
We need any solution to be cross platform (win/linux).

I'd like @supernothing to review this and add his thoughts.

Is there a way that we can have a separate service, running on the same host as polyswarm_client, that is monitoring polyswarm_client? Then we have a separate thread/process who's purpose is just to write a file with no content, but simply updating the file will update the timestamp, and then the external service will see the timestamp to know that the polyswarm_client service is working.

Rizato:
I wanted to do that as well, but the event loop running isn't enough. We have had issues in the past where the responses are blocked, but the event loop is still running.

Rizato:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command

PR #385, opened 2019-10-22
Add view-balance and view-stake sub commands in balancemanager

Rizato:


Rizato:
@mrtizmoatwork Can I cut a a minor version release with this?

PR #393, opened 2019-11-06
Create new session if in conn-refused loop

Rizato:
We are experience a case on stage where the websocket is connected and new bounties are coming in, but all http requests get connection refused. It can be fixed with a restart. 

This is attempting to trigger a new session, and see if that solves the issue without a restart


Rizato:
This doesn't seem to work at all.

PR #394, opened 2019-11-06
Just exit on connection refused or bad gateway

Rizato:
This is horrible. Engines will lose too much state due a polyswarmd outage, but less state than getting connection refuse looped forever, for whatever reason.


Rizato:
Not doing this.

PR #399, opened 2019-12-13
Settle all bounties for a contract on Deprecated event

Rizato:
When we are moving to a new contract, have all parties automatically get all their NCT from the contract.


Rizato:
Code climate warnings are about using md5 hashes in our metadata

PR #401, opened 2019-12-31
DN-3383

marciogualtieri-polyswarm:
DN-3383: Run micro-engine's blocking code inside an executor.

### DIFF
--- a/src/polyswarmclient/utils.py
+++ b/src/polyswarmclient/utils.py
@@ -169,7 +169,7 @@ def run_command_line(command):
     :param command: Command to run.
     :return: Command output or error message.
     """
-    process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
+    process = subprocess.Popen(command, stdout=subprocess.PIPE)
     process.wait()
     out, err = process.communicate()
     process.kill()
--- a/src/polyswarmclient/utils.py
+++ b/src/polyswarmclient/utils.py
@@ -4,7 +4,7 @@ import logging
 import os
 import sys
 import uuid
-from Crypto.Hash import keccak
+import sha3 as pysha3
 from concurrent.futures import ThreadPoolExecutor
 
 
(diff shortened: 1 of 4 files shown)


src/polyswarmclient/utils.py
```diff
@@ -1,20 +1,22 @@
 import asyncio
+import base58
```
alanjds-ps: Why was this moved up?

src/polyswarmclient/utils.py
```diff
@@ -139,3 +147,16 @@ def is_valid_ipfs_uri(ipfs_uri):
...
+def run_in_executor(f):
+    """
+    Decorator. Runs the function it decorates in an executor.
+    """
+    def inner(*args, **kwargs):
+        if sys.platform == 'win32':
+            loop = asyncio.ProactorEventLoop()
+        else:
+            loop = asyncio.get_event_loop()
+        return loop.run_in_executor(ThreadPoolExecutor(), lambda: f(*args, **kwargs))
```
alanjds-ps: Suggestion: If the loop can be got only once, can detect the platform outside `inner`.:
```py
if sys.platform == 'win32':
    loop = asyncio.ProactorEventLoop()
else:
    loop = asyncio.get_event_loop()

def inner(*args, **kwargs):
    return loop.run_in_executor(ThreadPoolExecutor(), lambda: f(*args, **kwargs))

return inner

File: src/polyswarmclient/utils.py [hunk 3]
```diff
@@ -139,3 +147,16 @@ def is_valid_ipfs_uri(ipfs_uri):
     except Exception as err:
         logger.exception('Unexpected error: %s', err)
     return False
+
+
+def run_in_executor(f):
+    """
+    Decorator. Runs the function it decorates in an executor.
+    """
+    def inner(*args, **kwargs):
```
alanjds-ps: Is a [good pratice](https://docs.python.org/3.8/library/functools.html#functools.wraps) to not lose [the metadata](https://stackoverflow.com/a/55102697/798575) of the decorated func when decorating it:
```suggestion
    @functools.wrap(f)
    def inner(*args, **kwargs):
```
(if accepting the suggestion, please remember to `import functools` ;) )

marciogualtieri-polyswarm:
Realized that I don't need this change anymore, but thank you for the review, @alanjds-ps.

PR #404, opened 2020-01-22
Fix windows tests

Rizato:


.gitlab-ci.yml
```diff
@@ -57,7 +57,7 @@ test-linux:
     - kube
   script:
     - docker pull $REPO_URL/$BASE_IMAGE_NAME:$CI_COMMIT_SHA
-    - docker run $REPO_URL/$BASE_IMAGE_NAME:$CI_COMMIT_SHA tox
+    - docker run $REPO_URL/$BASE_IMAGE_NAME:$CI_COMMIT_SHA pytest -s --cov=polyswarmclient
```
marciogualtieri-polyswarm: Tox creates a brand new python environment, so I believe that's the reason it works on build, but fails on test. Running pytest will fix the issue, but I think that adding `python-magic-bin==0.4.14;platform_system=="Windows"
python-magic==0.4.15;platform_system=="Linux"` to requirements.txt should fix it as well. I was not sure, so I created a branch to test it. :)

PR #409, opened 2020-01-29
Increase worker utilization defaults

Rizato:
Gets rid of a couple uses where Client and Worker had identical functions that started an infinite loop to update liveness. It now lives in the liveness class under `start`

Changes the worker flow to have a single task reading jobs from the queue, and then it spawns tasks to process it. 
Coupled with a change to defaults, now allows worker to create unlimited processing tasks, or scale as they see fit. 

Fixes a bug in ambassador


src/polyswarmclient/abstractambassador.py
```diff
@@ -185,90 +185,88 @@ def run(self):
...
-                                                              metadata=metadata)
-            await self.client.liveness_recorder.remove_waiting_task(bounty.ipfs_uri)
-            if not bounties:
-                await self.on_bounty_post_failed(bounty.artifact_type, bounty.amount, bounty.ipfs_uri, bounty.duration,
-                                                 chain, metadata=bounty.metadata)
+            if balance >= bounty.amount + bounty_fee:
+                return True
+            elif self.client.tx_error_fatal and tries >= MAX_TRIES:
+                logger.error('Failed %s attempts to post bounty due to low balance. Exiting', tries)
+                exit(1)
```
alanjds-ps: Backoff logic is always makes us sad.

Consider outsourcing it to [backoff](https://pypi.org/project/backoff/) lib

src/polyswarmclient/abstractambassador.py
```diff
@@ -185,90 +185,88 @@ def run(self):
...
 
-        logger.warning('Failed %s attempts to post bounty due to low balance. Skipping', tries, extra={'extra': bounty})
-        await self.on_bounty_post_failed(bounty.artifact_type, bounty.amount, bounty.ipfs_uri, bounty.duration, chain,
-                                         metadata=bounty.metadata)
+        assertion_reveal_window = await self.client.bounties.parameters[chain].get('assertion_reveal_window')
+        arbiter_vote_window = await self.client.bounties.parameters[chain].get('arbiter_vote_window')
+        metadata = None
+        if bounty.metadata is not None:
+            ipfs_hash = await self.client.bounties.post_metadata(bounty.metadata, chain)
+            metadata = ipfs_hash if ipfs_hash is not None else None
```
marciogualtieri-polyswarm: `metadata = ipfs_hash if ipfs_hash is not None else None` == `metadata = ipfs_hash`?

src/polyswarmclient/liveness/liveness.py
```diff
@@ -70,10 +80,11 @@ def __init__(self):
...
         """Mark a task as done processing"""
         async with self.waiting_lock:
-            try:
-                del self.waiting[key]
-            except KeyError:
-                logger.exception('%s not found in waiting tasks', key)
+            if key in self.waiting:
+                try:
+                    del self.waiting[key]
+                except KeyError:
```
marciogualtieri-polyswarm: Are you sure you need to log the exception? Doesn't the message give you all the information that is useful?

src/polyswarmclient/abstractambassador.py
Rizato: `self.client.tx_error_fatal` is set when the `Client` object is created

PR #412, opened 2020-01-30
Finalize 2.7.3

Rizato:


Rizato:
WIP so that we can wait on CI for windows engines so I can test

PR #415, opened 2020-02-05
Update to aiohttp 3.6.2

Rizato:
Includes a fix for stream responses, which will help out some engines


Rizato:
Putting to WIP so I can change engines that speify 3.5.4 to use whatever polyswarm-client gives them.

PR #433, opened 2020-05-19
Add `TemporaryArtifactFile` async context manager

zv:


src/polyswarmclient/abstractscanner.py
```diff
@@ -1,10 +1,58 @@
...
+            logger.info('wrote %d bytes to %s', written, f.name)
+            yield f
+        finally:
+            await loop.run_in_executor(None, f.close)
+
+else:
+    from async_generator import async_generator, yield_, asynccontextmanager
+
+    @asynccontextmanager
+    @async_generator
```
Rizato, line 38: I don't think we need this with 3.6

src/polyswarmclient/abstractscanner.py
```diff
@@ -1,10 +1,58 @@
...
+            f = tempfile.NamedTemporaryFile(**{
+                'mode': 'w+b',
+                'prefix': 'polyswarm-artifact',
+                'buffering': 0,
+                'delete': True,
+                **kwargs
+            })
+            written = await loop.run_in_executor(None, f.write, blob)
+            logger.info('wrote %d bytes to %s', written, f.name)
+            await yield_(f)
```
Rizato, line 51: I think this can be a normal yield in 3.6

src/polyswarmclient/abstractscanner.py
Rizato: Just the generator decorator

PR #438, opened 2020-05-29
Allow synchronous scanners

Rizato:
Provides `platform.system()` and `platform.machine()` values as fields in AbstractScanner. 

We should only need to call these once when the scanner starts, instead of on every scan to build metadata.


Rizato:
Cancelled CI on 30d5eaf and d6a6be1 because of later commits

Rizato:
Let me know if there is a known pattern or better way to identify which function it should call.

zv:
blammo!

src/polyswarmclient/abstractscanner.py
```diff
@@ -28,14 +31,65 @@ def __repr__(self):
...
+    This class offers two scan options, which can be specified by passing a ScanMode enum value as `mode`.
+    It uses asynchronous scan by default.
+
+    The function `scan_async` is a coroutine function where everything called from this function must be async compatible. non-blocking,
+    That means it uses only non-blocking IO, and runs nothing cpu-bound, like hash functions.
+
+    The function `scan_sync` is a synchronous function where anything goes.
+    It is called in a ThreadPoolExecutor so it is compatible with the asynch code in the worker.
+
+
```
alanjds, line 55:
```suggestion
```

src/polyswarmclient/abstractscanner.py
```diff
@@ -46,15 +100,14 @@ async def scan(self, guid, artifact_type, content, metadata, chain):
...
         """
-        pass
+        raise NotImplementedError("Must implement scan_async when using ScanMode.Async")
 
     async def setup(self):
         """Override this method to implement custom setup logic.
 
         This is run immediately after the Scanner class is instantiated and before any calls to the scan() method.
 
         Args:
```
alanjds, line 110: Consider to remove docstrings on `Args:` if no arg will be described.
```suggestion
        This is run immediately after the Scanner class is instantiated and before any calls to the scan() method.
```

PR #439, opened 2020-06-05
Add scan time requirement

Rizato:
Let's us set the requirement remaining time to complete a scan
Any scan with a `remaining_time` lower than this requirement raises `ExpiredException`


Rizato:
Set the default to 3 per @scsich request.

src/worker/__main__.py
```diff
@@ -71,7 +71,9 @@ def choose_backend(backend):
               help='Activate testing mode for integration testing, respond to N bounties and N offers then exit')
 @click.option('--log-format', default='text',
               help='Log format. Can be `json` or `text` (default)')
-def main(log, client_log, redis_addr, queue, backend, tasks, download_limit, scan_limit, api_key, testing, log_format):
+@click.option('--scan_time_requirement', envvar='SCAN_TIME_REQUIREMENT', default=3,
+              help="A minimum time it takes to scan a file in order to cull impossible scans")
```
mrtizmoatwork, line 75: In the help msg, you should specify that this is in seconds

mrtizmoatwork: "Minimum number of seconds this engine requires to scan a file; artifacts with less time remaining will be ignored."

PR #440, opened 2020-06-05
Use slim-stretch

Rizato:


mrtizmoatwork:
need to add a step to install these deb's:
- wget
- unzip

```
RUN apt-get update && apt-get install -y \
	unzip \
	wget \
  && rm -rf /var/lib/apt/lists/*
```

PR #441, opened 2020-06-06
Update source image to `python:3.7-slim-buster`

zv:
This commit also installs a minimal set of packages used in later image builds (e.g `gcc`, `make`, `pkg-config`, etc.)

The image is less than half the size of the existing image:

```console
zv@ps:polyswarm-client$ docker image ls | grep polyswarm-client | head -n 2
zv/polyswarm-client		latest	7f7a9a53085f	7 minutes ago	468MB
polyswarm/polyswarm-client	latest 	380d136725fb	3 weeks ago	1.03GB
```

The `Dockerfile` also has new `CFLAGS`, `CXXFLAGS` and `LDFLAGS` environment variables common to each of the instance types we currently use in EC2 (i.e Haskell & Skylake), although if people are downloading this image from `Dockerhub`, we should bring this back a couple of generations or remove it entirely.


requirements.txt
```diff
@@ -2,29 +2,29 @@ async-timeout<4.0,>=3.0
...
 git+https://github.com/polyswarm/python-clamd.git@async#egg=clamd
-click>=6.7
-coverage==4.5.1
+click>=7.1.2
+coverage==5.1
 dataclasses==0.7; python_version == '3.6'
-jsonschema==3.0.2
-hypothesis==3.82.1
-tox==3.4.0
+jsonschema==3.2.0
```
Rizato, line 14: Did the polyswarm-artifact changes get pushed in a new release yet?

mrtizmoatwork:
We should put the CFLAGS stuff in a separate Dockerfile, so we don't break external parties' use.

So, something like this:
1. First Dockerfile, do the base build and push to Dockerhub and internal ECR (polyswarm-client)
2. Second Dockerfile, add our internal linux tweaks and push to internal ECR (polyswarm-client-internal-linux)
3. Third Dockerfile, add our internal windows tweaks and push to internal ECR (polyswarm-client-internal-windows)

PR #444, opened 2020-06-12
Restore external clients

Rizato:
Supports current polyswarmd on prod communities. 

Reminder: Merge this one, not squash, since it contains a bunch of squash commits, and the history is important since each commit is a feature.


src/polyswarmclient/__init__.py
```diff
@@ -136,46 +121,100 @@ async def run_task(self, chains=None, listen_for_events=True):
...
+    def create_fast_sub_clients(self, chains):
+        from polyswarmclient.fast import BalanceClient,  BountiesClient, StakingClient, OffersClient, RelayClient
+        self.bounties = BountiesClient(self)
+        self.staking = StakingClient(self)
+        self.offers = OffersClient(self)
+        self.relay = RelayClient(self)
+        self.balances = BalanceClient(self)
+
+        async def periodic(chains):
+            # FIXME PSC continues to hit a down polyswarmd, because the trigger is time, not blocks from websocket
```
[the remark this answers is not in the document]
Rizato: I'm not sure how to fix this one yet.

requirements.txt
```diff
@@ -1,22 +1,21 @@
...
-aioresponses==0.6.3
+aioresponses==0.6.4
 aiorwlock==0.6.0
 backoff==1.10.0
 base58==2.0.0
 cachetools==4.1.0
 git+https://github.com/polyswarm/python-clamd.git@async#egg=clamd
-click>=6.7
-dataclasses==0.7; python_version == '3.6'
+click==7.1.2
```
Rizato, line 11: Does e2e support this? Last I checked we have to be at >=6.7

requirements.txt
```diff
@@ -1,22 +1,21 @@
...
-python-magic==0.4.15;platform_system=="Linux"
+python-json-logger==0.1.9
 python-magic-bin==0.4.14;platform_system=="Windows"
-requests==2.23.0
-web3==5.6.0
-websockets==8.1
-yara-python==3.11.0
+python-magic==0.4.15;platform_system=="Linux"
+requests==2.22.0
+web3==4.8.2
```
Rizato: I think we want to be on later versions of web3.

Rizato:
I'm good with the current code climate issues.

tests/utils/mockups.py
```diff
@@ -0,0 +1,141 @@
...
+        logger.info("Loading Mock scanner...")
+        if artifact_types is None:
+            artifact_types = [ArtifactType.FILE, ArtifactType.URL]
+        if scanner is None:
+            scanner = MockScanner()
+
+        super().__init__(client, testing, scanner, chains, artifact_types)
+
+
+with tempfile.NamedTemporaryFile(delete=True) as key_file:
```
zv, line 121: If it's your ultimate ambition to turn `mock_ambassador`, `mock_microengine`, etc into objects which can be fed into test functions, I **highly** recommend looking into [`pytest` fixtures](https://docs.pytest.org/en/stable/fixture.html) before you do.

`pytest` even has some builtin fixtures for temporary file creation which might justify the cost of admission alone, e.g:

```python
...
import pytest
...

@pytest.fixture(params=[(TESTKEY, TESTKEY_PASSWORD)])
def keyfile(key, password, tmp_path):
    kf = tmp_path / 'keyfile'
    kf.write_bytes(key)
    return (str(kf), password)

@pytest.fixture(scope='session')
def mock_ambassador(keyfile):
    return MockAmbassador.connect(
        POLYSWARMD_FAST_ADDRESS,
        *keyfile, # keyfile is a tuple of (FILENAME, PASSWORD)
        insecure_transport=True,
        chains=CHAINS,
    )
```

src/worker/__init__.py
```diff
@@ -96,15 +98,19 @@ def start(self, loop: asyncio.AbstractEventLoop):
...
-        self.setup_synchronization(loop)
+        self.setup_process_pool()
+        self.setup_semaphores(loop)
         self.setup_graceful_shutdown(loop)
         await self.setup_liveness(loop)
         await self.setup_redis(loop)
         if not await self.scanner.setup():
             logger.critical('Scanner instance reported unsuccessful setup. Exiting.')
-            exit(1)
+            raise FatalError('Scanner setup failed', 1)
```
zv, line 108: You're using `FatalError` in places outside of CLI interface modules, we probably shouldn't make internals code dependent on `click`.

src/polyswarmclient/abstractambassador.py
```diff
@@ -197,30 +197,30 @@ async def submit_bounty(self, bounty, chain):
...
+            bounty_fee = await self.client.bounties.parameters[chain].get('bounty_fee')
+            try:
+                await self.client.balances.raise_for_low_balance(bounty.amount + bounty_fee, chain)
+            except LowBalanceError as e:
+                await self.on_bounty_post_failed(bounty.artifact_type, bounty.amount, bounty.ipfs_uri, bounty.duration,
+                                                 chain, metadata=bounty.metadata)
+                self.bounty_queues[chain].task_done()
+                self.bounty_semaphores[chain].release()
+                if self.client.tx_error_fatal:
+                    raise FatalError('Failed to post bounty due to low balance') from e
```
zv, line 210: Same story, we probably shouldn't make internals code dependent on `click`. Consider raising a different exception entirely which does something similar.

src/polyswarmclient/abstractambassador.py
```diff
@@ -326,7 +326,7 @@ async def __handle_new_block(self, number, chain):
             bounties_posted = self.bounties_posted.get(chain, 0)
             last_bounty_count = self.last_bounty_count.get(chain, 0)
             if blocks % self.watchdog == 0 and bounties_posted == last_bounty_count:
-                raise Exception('Bounties not processing, exiting with failure')
+                raise FatalError('Bounties not processing')
```
zv, line 329: Same here, I think we probably want to create and raise a totally new exception here.

src/polyswarmclient/abstractmicroengine.py
```diff
@@ -229,51 +228,46 @@ async def __handle_new_bounty(self, guid, artifact_type, author, amount, uri, ex
...
+            assertion_fee = await self.client.bounties.parameters[chain].get('assertion_fee')
+            assertion_reveal_window = await self.client.bounties.parameters[chain].get('assertion_reveal_window')
+            arbiter_vote_window = await self.client.bounties.parameters[chain].get('arbiter_vote_window')
+
+            bid = await self.bid(guid, mask, verdicts, confidences, metadatas, chain)
+            try:
+                await self.client.balances.raise_for_low_balance(assertion_fee + sum(bid), chain)
+            except LowBalanceError as e:
+                if self.client.tx_error_fatal:
+                    raise FatalError('Failed to assert on bounty due to low balance') from e
```
zv, line 257: Another `FatalError` - we could probably just use something like `FatalError` here that doesn't inherit from `ClickException`.

src/polyswarmclient/exceptions.py
```diff
@@ -25,3 +27,49 @@ class LowBalanceError(PolyswarmClientException):
...
+    pass
+
+
+class UnsupportedHashError(PolyswarmClientException):
+    """
+    Raised when a hash doesn't match the format of a hash we use
+    """
+
+
+class FatalError(ClickException):
```
zv, line 72: I like the idea of `FatalError` but I'm not a huge fan of `FatalError` inheriting from `ClickException`.

Rizato: All of those errors are something the user has to see. ClickExceptions show the user the error in the nice click way. 


I am pretty sure we only use it in classes that have a command line tool associated with them. microengine, arbiter, ambassador, and worker are all intended to be run from click, so it makes sense to me that fatal errors in those classes should be shown in the cli.

tests/utils/mockups.py
Rizato: Yea, agreed. Those should have been in fixtures. 

They aren't used anywhere anymore anyway, they just supported tests that didn't test anything and got removed a while ago. Just removing this file.

src/polyswarmclient/ethereum/transaction/base.py
```diff
@@ -0,0 +1,273 @@
...
+logger = logging.getLogger(__name__)
+
+LOG_MSG_ENGINE_TOO_SLOW = ('PLEASE REVIEW YOUR SCANNING LOGIC. '
+                           'Bounty inactive errors indicate that the microengine received the bounty, '
+                           'but was unable to respond to the bounty within the time window. '
+                           'Such errors are considered fatal during testing so you can easily identify them. '
+                           'If your engine is unable to respond within the time window on the live PolySwarm '
+                           'network, you risk losing the bid amount of the bounty at hand. We strongly '
+                           'encourage you to review your artifact scan process to identify areas where engine '
+                           'speed can be improved.')
```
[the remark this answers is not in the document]
Rizato: I think it was done that way for easier change tracking with git. Each sentence on a separate line, just like markdown. 

All I did was move this code, so I'd prefer to leave as is.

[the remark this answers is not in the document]
Rizato: Nevermind, it's not each sentence on a different line.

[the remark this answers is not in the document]
Rizato: Okay, I updated it to use triple quotes, and be one sentence per line. Leaving out the replace, it looked fine when printed.

src/polyswarmclient/exceptions.py
Rizato: I resolved the others, so we can just chat in this one easily.

PR #447, opened 2020-06-16
Fix merge problems

Rizato:
Fix some issues with resolving divergent worker implementations. 

Trying to revert to working code.


Rizato:
This is finally ready to go.No more caught errors when running.

PR #452, opened 2020-06-23
Allow non-default executor for scan_sync

Rizato:
Allows developers the option of choosing their `Executor` implementation, while providing a good default. 

Developers will know their use case better, and may want to use the `ThreadPoolExecutor` if their `scan_sync` implementation relies heavily on IO, or stay with the `ProcessPoolExecutor` if they are using cpu bound operations.


tests/test_scanner.py
```diff
@@ -1,68 +1,102 @@
...
+    def scan_sync(self, guid, artifact_type, content, metadata, chain):
+        time.sleep(self.sleep)
+        return ScanResult(verdict=True)
+
+    async def scan_async(self, guid, artifact_type, content, metadata, chain):
+        time.sleep(self.sleep)
+        return ScanResult(verdict=False)
+
+    def get_executor(self) -> Optional[Executor]:
+        return ProcessPoolExecutor(max_workers=1)
```
Rizato, line 57: I would have likes to have this in a fixture as well, but pickle complained so much.

Rizato:
So the answer was, ProcessPoolExecutor in a fixture, does not work. At least on windows. 
Just moved it into a separate test, same code otherwise.

You can see all my other attempts.

PR #454, opened 2020-06-24
Add a daily rate limit option to the backend

Rizato:
This will only increment the rate limit when a file is scanned. 
Expired bounties are not counted against the rate limit when used in the backend, unlike in the producer. 

Uses a different key so both rate limits can be in use, though is not recommended.


mrtizmoatwork:
please change log messages to not use f-strings.

Rizato:
Test in e2e, and it does stop the worker after the limit is reached.

PR #456, opened 2020-06-30
Read JobResponse objects from redis queues in a single task

Rizato:
Stops calling `wait_for_results()` for every bounty, and instead register jobs with the new `JobProcessor` class. 

The `JobProcessor` kicks off a task to get all results for all jobs every .5 seconds, and uses a `Future` object to return results to the waiting task. 

The `JobProcessor` does all the expired job handling, confidence modifier requests, and aggregates ScanResult in a backwards compatible fasion, so neither the Arbiter, or Microengine `producer` backend options need to be changed.


Rizato:
I think I need to find a solution with less locked time. I can imagine the list getting updated gets too large, and it becomes impossible to get a new task in before it is expired.

PR #458, opened 2020-07-01
Add variable to skip E2E

Rizato:
Let developers skip e2e by setting SKIP_E2E variable in ci


Rizato:
Ran the latter pipeline with SKIP_E2E and you can confirm that it did skip e2e

Rizato:
Re running the OG pipeline, so you can see it both ways

PR #460, opened 2020-07-06
Use a single redis pool per producer/worker

Rizato:
Windows engines are sensitive to the amount of connections to redis. Even when they are from linux engines. This reduces the number of connections a bit. 

This also makes sure any `RedisDailyRateLimit` objects use the same connection pool as the producer or worker


tests/utils/fixtures.py
```diff
@@ -223,6 +224,22 @@ def mock_client(event_loop):
...
+@pytest.fixture()
+def redis_uri():
+    return 'redis://redis:6379'
+
+
+@pytest.fixture()
+@pytest.mark.asyncio
+async def redis_client(event_loop):
+    redis = await aioredis.create_redis_pool('redis://redis:6379')
+    with await redis as redis_client:
```
alanjds, line 236: Is strange to read `with async` instead of `async with`

Rizato: Yea. It's a weird part of aioredis. 

It doesn't define `__aenter__`, `__aexit__`

It defines `__await__` which returns a ContextRedis object that implements `__enter__`, and `__exit__`

Rizato: @alanjds

Rizato: Behind the scenes it uses Threads and a `Future` object to make it async compatible

PR #464, opened 2020-07-17
Use backoff in infinite loop

Rizato:
Backoff as a decorator is meant for functions that return. Each time the function is called, it gets a new sleep time generator. 
When used in an infinite loop, the sleep time grows until it reaches the max, and stays there forever. 

This adds a wrapper class called `BackoffWrapper` that takes a `backoff` generator function, and lets an infinite loop use it. 
To use it, create a `BackoffWrapper` object. Create the infinite loop, and wrap the code with a try-except. 

Every time the task being performed works, call `reset()`. In the except blocks call `sleep()`. Every time sleep is called without a reset, the backoff will grow. 

This is not as clean as a decorator with all of the exceptions at the top, but does work with infinite loops

In addition there are a couple other changes in this PR. 

1. separates `listen_for_events` into three functions for readability
1. removes the backoff for `Client.run` and `Worker.run`. It now exits on unexpected errors. This is a breaking change.


Rizato:
Yea, I went to their issues where people wanted the context manager option. But it really doesn't work.

PR #465, opened 2020-07-24
Stop exiting the websocket loop

Rizato:


Rizato:
Closing in favor of #466

PR #466, opened 2020-07-24
Expire and finish PendingJobs in `fetch_results`

Rizato:
The previous JobProcessor fetched results in 3 steps. First it expired old jobs, then updated all jobs, and finally, let the tasks waiting on the assertion/vote move forward after all jobs were updated.

 This is not ideal, especially with some 250 simultaneous bounties all trying to hit redis, and some job has to wait to assert for all the others. 

The updated version, does the expiration and done checks in `fetch_results`. Then `fetch_results` will set the value in the `Future`, immediately allowing the assertion/vote to be sent to `polyswarmd`. 

This also reduces the amount of locking so more things can run at once.


Rizato:
This one catches quite a few changes in it since we are changing our release schedule to release on master merge.

Rizato:
> : I am curious about why lots of the API got true-private on this PR.

A few reasons, @alanjds 

1. I think I finally got over my hatred of the `__name` syntax for private functions. At least, partially. I would still prefer the word private
1. The changes to PendingJob made it so most of those methods don't need to be exposed anymore, since the object handles it's operations itself, now. 
1. For some, I just moved into separate methods for readability, but I don't them to be called. 
1. Since the JobProcessor/PendingJob weren't in previous releases, it won't break anything to change it now.

PR #469, opened 2020-08-11
Allows api keys to be sent over HTTP

Rizato:
Deprecates `--insecure-transport` in favor of `--allow-keys-over-http`. 

Use of this key flag is not recommended as third parties may gain access to key information. However, there are legitimate business applications where http is used, but the key material is safe. 

Again, use of this flag is not recommended.


Rizato:
Last change. Sorry @alanjds for all the reviews.

PR #473, opened 2020-10-12
Set log level and format via env var

Rizato:
Working towards setting up a polyswarm-client in a helm chart without needing to overwrite command


mrtizmoatwork (approved):
ok

PR #475, opened 2020-10-29
Add option to rate limit via redis based on a daily limit, down to secondly. 

Rizato:
Takes a kwarg `accuracy` to determine the keyname, and expiration time.

Provides default accuracy options of Day, Hour, Minute, Second.
Default accuracy is Day.

This way engines can fine tune more rate limits in their `Scanner` impl beyond just the daily limit exposed as a `Worker` command line argument.


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.10.3
+current_version = 2.11.0
```
Rizato, line 2: Should this be 10.4 or 11.0?

zv: 11.0 is right.

src/polyswarmclient/ratelimit/redis.py
```diff
@@ -1,34 +1,82 @@
...
+    @abstractmethod
+    def get_key(self, prefix):
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_expiration(self):
+        raise NotImplementedError()
+
+
+class DateKeyManager(RedisKeyManager):
```
zv, line 29: I was initially skeptical of using a subclass-per-delta here, but after some thought I really like this.

src/polyswarmclient/ratelimit/redis.py
```diff
@@ -1,34 +1,82 @@
...
+
+
+class MinutelyKeyManager(DateKeyManager):
+    def __init__(self):
+        super().__init__(expiration=MINUTELY_SECONDS, format_str='%Y-%m-%d-%H-%M')
+
+
+class SecondlyKeyManager(DateKeyManager):
+    def __init__(self):
+        super().__init__(expiration=SECONDS, format_str='%Y-%m-%d-%H-%M-%S')
```
zv, line 59: Should this `format_str` be `%Y-%m-%d %H:%M:%S`?

src/polyswarmclient/ratelimit/redis.py
```diff
@@ -1,34 +1,82 @@
...
+
+
+class HourlyKeyManager(DateKeyManager):
+    def __init__(self):
+        super().__init__(expiration=HOURLY_SECONDS, format_str='%Y-%m-%d-%H')
+
+
+class MinutelyKeyManager(DateKeyManager):
+    def __init__(self):
+        super().__init__(expiration=MINUTELY_SECONDS, format_str='%Y-%m-%d-%H-%M')
```
zv, line 54: Should this `format_str` be `%Y-%m-%d %H:%M`?

src/polyswarmclient/ratelimit/redis.py
```diff
@@ -1,34 +1,82 @@
...
+        raise NotImplementedError()
+
+
+class DateKeyManager(RedisKeyManager):
+    def __init__(self, expiration, format_str):
+        self.expiration = expiration
+        self.format_str = format_str
+
+    def get_key(self, prefix):
+        date = datetime.date.today().strftime(self.format_str)
```
zv, line 35: `datetime.date` only capture the date, so time fields will all be zero.

I think we'll be better served by `datetime.datetime.today()` or even `datetime.datetime.utcnow()`

```
>>> import datetime 
>>> datetime.date.today().strftime('%Y-%m-%d-%H-%M-%S')
2020-10-30-00-00-00
>>> datetime.datetime.today().strftime('%Y-%m-%d-%H-%M-%S')                 
2020-10-30-13-06-35
```

zv (changes requested):

src/polyswarmclient/ratelimit/redis.py
```diff
@@ -1,34 +1,82 @@
...
     """ Third Party limitation where redis is used to track a daily scan limit.
         Keys are based on the current date, and will expire the next day.
 
         This implementation is used in the producer and worker since they use Redis already.
     """
-    def __init__(self, redis, queue, limit):
+    def __init__(self, redis, queue, limit, key_manager=None):
         self.redis = redis
         self.queue = queue
         self.limit = limit if limit is None else int(limit)
```
zv, line 71: I know you didn't add this line, but I'm pretty sure you can just set `limit=None` as a default argument, with no funky business occurring (I believe `None` is a value, not reference)

zv (approved):

PR #476, opened 2020-11-02
Add options to enable more precise rate limits in worker

Rizato:
Add `worker` command options to use hourly, minutely, and secondly rate limits in worker. 

Originally, these features were planned to be used in `Scanner` implementations, but the `Scanner` doesn't have all the information (redis connection, and queue name) to use the existing infrastructure for rate limits. 

This adds `--hourly-rate-limit`, `--minutely-rate-limit`, and `--secondly-rate-limit` options to the  `worker` command. 

I considered creating a new `--rate-limit` command in the same vein as `--filter` with the ability to create multiple commands with a full structure to determine the timing and limits. However, we need to keep `--daily-rate-limit` for backwards compatibility, and both commands would conflict with the zen of python rule `There should be one-- and preferably only one --obvious way to do it.`


Rizato:
Adding tests for the aggregate class

zv (approved):

zv:
I don't think it's worth changing at this point, but our next rehash of this should allow us to somehow get at the length of the time period each `KeyManager` corresponds to, maybe just expressing each as some time N mod SECONDS == 0.

Rizato:
When the webhook stuff launches, I would like to change these commands up, and move to a `--filter` style command for it. 

For, letting the rate limit work for periods that aren't per second/minute/hour/day we will have to create a new key scheme. We are at the limit of just using a string format as the key in my mind.

PR #480, opened 2020-11-20
Update counters in transaction

Rizato:
Due to restarts or errors, these counters can get out of sync. 

This change runs the whole group as a single transaction so either all or none of them update.


mrtizmoatwork (approved):
ok

mrtizmoatwork (approved):
ok

PR #482, opened 2020-12-23
Resolve memory leak by removing pending_bounties

Rizato:
Removes `pending_bounties` from abstractarbiter and abstractmicroengine.

`pending_bounites` was intended to deduplicate bounties due having multiple blockchains nodes.
Deduplication was important to avoid rescanning a file again, which would add roadblocks to scanning new bounties in time.
It was also important to stop wasting transactions with duplicate settle attempts.
    
Originally, an item was added to `pending_bounites`, and removed upon settle.
In our current system, where settles are temporarily disabled, `pending_bounties` grows forever.
    
We long ago solved the duplication issue, and settles will no longer be the microengine's responsibility.
`pending_bounites` is no longer necessary, and was honestly always doomed to fail when we increased the settle time to 7 days.


mrtizmoatwork (approved):
ok

alanjds (dismissed):

mrtizmoatwork:
hey @Rizato while you are at it, can you update the `.gitlab-ci.yaml` to use the new runners?

Rizato:
@mrtizmoatwork What should I be changing for that?

Rizato:
Found the wiki

mrtizmoatwork (approved):
ok

PR #483, opened 2020-12-23
remove consumer as a downstream project. we no longer use it.

mrtizmoatwork:


Rizato (approved):

alanjds (approved):
We could think about using a DAG if this gets needed again in the future...

PR #490, opened 2021-03-30
Fix missing $

Rizato:
Missed the dollar sign on some of the tags causing errors.


alanjds (dismissed):

Rizato:
Bumping the version because 2.11.7 made it to pypi, but not docker.

mrtizmoatwork (approved):
ok

PR #495, opened 2021-04-16
Pass full url to worker

Rizato:
Constructs the artifact url in the frontend. 

worker just downloads from the url it is given, (rather than constructing from the parts).


alanjds (approved):

Rizato:
Tested against e2e at master-fast. Tested against e2e at master.

alanjds (approved):

alanjds (approved):

PR #499, opened 2021-05-04
Revert "Pass image tag downstream"

Rizato:
Reverts polyswarm/polyswarm-client#493


Rizato:
Going to have one commit that undoes all the changes that need to be undone.

PR #500, opened 2021-05-05
Revert changes for planned breaking changes. 

Rizato:
This reverts a few commints, and upgrades pycryptodome to fix tests.


mrtizmoatwork (approved):
ok

Rizato:
Ended up testing with orchestration. Glad we fixed that a while back.

PR #503, opened 2021-08-03
Fix hanging worker after Redis failure

Rizato:
Updates to aioredis 2.0.0. 
Uses a similar API to redis-py, but async. 

Catches TimeoutError in worker, which causes the worker to hang when there was a Timeout talking to redis.


src/worker/base.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, redis_addr, queue, task_count=0, download_limit=0, scan_limit
                  scanner=None, scan_time_requirement=0, daily_rate_limit=None, hourly_rate_limit=None,
                  minutely_rate_limit=None, secondly_rate_limit=None, allow_key_over_http=False):
         self.redis_uri = 'redis://' + redis_addr
-        self.redis = None
+        self.redis = aioredis.from_url(self.redis_uri)
```
alanjds, line 93: So the redis setup now takes place on `__init__` instead of `setup()`. Alright.

alanjds (approved):
Yeah, feels good to reuse the same API of a better-known flavor of a Redis lib.

src/worker/base.py
Rizato: I should probably check on this.

We may need to create this when we know the event loop has been set. AKA where the windows or linux event loop has been created.

alanjds (approved):
Pipelines are nice.

Rizato:
I'm realizing this is a breaking changes for any Engines that use this. Especially if they use Redis. Going to close this, and open a new one that just fixes the TimeoutError.

